PR #11, opened 2025-03-18
Configure build using UBI

gbenhaim:


Dockerfile
```diff
@@ -1,9 +1,10 @@
 # Build the manager binary
-FROM docker.io/golang:1.23 AS builder
+FROM registry.access.redhat.com/ubi9/go-toolset:9.5-1742197705 AS builder
```
filariow, line 2: I'd suggest to use sha instead of tag
```suggestion
FROM registry.access.redhat.com/ubi9/go-toolset@sha256:564c50dbad93a50dfe1439b295f021dae0bdc8c2aef3ad8be7b2a4dde52f0e2f AS builder
```

filariow (changes requested):

PR #21, opened 2025-04-24
Add integration tests and README

gbenhaim:
1. Add integration test
    2. Update the README with the project's information.

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -2,8 +2,6 @@ name: E2E Tests
 
 on:
   pull_request:
-    branches-ignore: 
-      - "*"
 
 jobs:
   test-e2e:
@@ -29,8 +27,19 @@ jobs:
 
       - name: Create kind cluster
         run: kind create cluster
+        env:
+          KIND_EXPERIMENTAL_PROVIDER: podman        
 
       - name: Running Test e2e
+        env:
+          KIND_EXPERIMENTAL_PROVIDER: podman
+          IMG: konflux-ci.dev/tekton-kueue:v0.0.1
         run: |
           go mod tidy
-          make test-e2e
+          make load-image "IMG=$IMG"
+          podman exec kind-control-plane crictl images
+          make kueue
+          make tekton
+          ret=0
+          make test-e2e || ret=$?
+          kubectl get pods -n tekton-kueue-system
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -40,6 +40,4 @@ jobs:
           podman exec kind-control-plane crictl images
           make kueue
           make tekton
-          ret=0
-          make test-e2e || ret=$?
-          kubectl get pods -n tekton-kueue-system
+          make test-e2e
--- a/Makefile
+++ b/Makefile
@@ -251,4 +251,4 @@ cert-manager-undeploy:
 
 .PHONY: load-image
 load-image: docker-build
-	dir=$$(mktemp -d) && podman save $(IMG) -o $${dir}/tekton-kueue.tar && kind load image-archive -n $(KIND_CLUSTER)  $${dir}/tekton-kueue.tar
+	dir=$$(mktemp -d) && $(CONTAINER_TOOL) save $(IMG) -o $${dir}/tekton-kueue.tar && kind load image-archive -n $(KIND_CLUSTER)  $${dir}/tekton-kueue.tar
--- a/Makefile
+++ b/Makefile
@@ -251,4 +251,7 @@ cert-manager-undeploy:
 
 .PHONY: load-image
 load-image: docker-build
-	dir=$$(mktemp -d) && $(CONTAINER_TOOL) save $(IMG) -o $${dir}/tekton-kueue.tar && kind load image-archive -n $(KIND_CLUSTER)  $${dir}/tekton-kueue.tar
+	dir=$$(mktemp -d) && \
+	$(CONTAINER_TOOL) save $(IMG) -o $${dir}/tekton-kueue.tar && \
+	kind load image-archive -n $(KIND_CLUSTER) $${dir}/tekton-kueue.tar && \
+	rm -r $${dir}
--- a/Makefile
+++ b/Makefile
@@ -143,7 +143,7 @@ endif
 
 .PHONY: install
 install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
-	[[ ! -d config/crd ]] && { echo "config/crd directory doesn't exist" || true; } && exit 1
+	[[ ! -d config/crd ]] && { echo "config/crd directory doesn't exist"; } || \
 	$(KUSTOMIZE) build config/crd | $(KUBECTL) apply --server-side -f -
 
 .PHONY: uninstall
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 # Image URL to use all building/pushing image targets
-IMG ?= controller:latest
+IMG ?= tekton-kueue:latest
+KIND_CLUSTER ?= kind
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
@@ -12,7 +13,7 @@ endif
 # Be aware that the target commands are only tested with Docker which is
 # scaffolded by default. However, you might want to replace it to use other
 # tools. (i.e. podman)
-CONTAINER_TOOL ?= docker
+CONTAINER_TOOL ?= podman
 
 # Setting SHELL to bash allows bash commands to be executed by recipes.
 # Options are set to exit when a recipe line exits non-zero or a piped command fails.
@@ -72,7 +73,7 @@ test-e2e: manifests generate fmt vet ## Run the e2e tests. Expected an isolated 
 		echo "Kind is not installed. Please install Kind manually."; \
 		exit 1; \
 	}
-	@kind get clusters | grep -q 'kind' || { \
+	@kind get clusters | grep -q -E 'kind|kueue' || { \
 		echo "No Kind cluster is running. Please start a Kind cluster before running the e2e tests."; \
 		exit 1; \
 	}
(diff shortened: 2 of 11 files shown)


config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: localhost/tekton-kueue
-  newTag: v2
+  newName: quay.io/konflux-ci/tekton-kueue
```
filariow, line 7: should we avoid to have a default registry here?

```suggestion
  newName: konflux-ci/tekton-kueue
```

test/e2e/e2e_test.go
```diff
@@ -66,9 +87,40 @@ var _ = Describe("Manager", Ordered, func() {
...
+		By(fmt.Sprintf("Creating a namespace: %s", nsName), func() {
+			ns := &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: nsName,
+				},
+			}
+			err := k8sClient.Create(ctx, ns)
+			if !(err == nil || kerrors.IsAlreadyExists(err)) {
+				Expect(err).ToNot(HaveOccurred())
+			}
```
filariow, line 108: If you like it, we could rely on `Satisfy` here

```suggestion
			Expect(k8sClient.Create(ctx, ns)).To(Satisfy(func(err error) bool {
				return err == nil || kerrors.IsAlreadyExists(err)
			}
```

test/e2e/e2e_test.go
```diff
@@ -66,9 +87,40 @@ var _ = Describe("Manager", Ordered, func() {
...
+				"kubectl",
+				"apply",
+				"--server-side",
+				"-n",
+				nsName,
+				"-f",
+				"config/samples/kueue/kueue-resources.yaml",
+			)
+			_, err := utils.Run(cmd)
+			Expect(err).NotTo(HaveOccurred(), "Failed to apply kueue resources")
```
filariow, line 122: something like this should be possible too

```suggestion
			Expect(utils.Run(cmd)).To(Succeed(), "Failed to apply kueue resources")
```

test/e2e/e2e_test.go
```diff
@@ -296,8 +387,193 @@ var _ = Describe("Manager", Ordered, func() {
 		//    strings.ToLower(<Kind>),
 		// ))
 	})
+
+	Context("N pipelines complete successfully", func() {
+		// todo generate a new one
```
filariow, line 392: what does it mean?

test/e2e/e2e_test.go
```diff
@@ -296,8 +387,193 @@ var _ = Describe("Manager", Ordered, func() {
 		//    strings.ToLower(<Kind>),
 		// ))
 	})
+
+	Context("N pipelines complete successfully", func() {
```
filariow, line 391: I think we should mark this as `Ordered` (cf. https://onsi.github.io/ginkgo/#ordered-containers)

> Ordered is a decorator that allows you to mark a container as ordered.  Specs in the container will always run in the order they appear.
They will never be randomized and they will never run in parallel with one another, though they may run in parallel with other specs.


```suggestion
	Context("N pipelines complete successfully", Ordered, func() {
```

filariow, line 391: should we have an `AfterAll` function that tries to delete all the PipelineRuns?

test/e2e/e2e_test.go
```diff
@@ -296,8 +387,193 @@ var _ = Describe("Manager", Ordered, func() {
...
+					return nil
+				},
+					(15*plrCount)*int(time.Second),
+					3*time.Second,
+				).Should(Succeed())
+			}
+		})
+	})
+
+	Context("Pipeline is queued when resources are missing", func() {
```
filariow, line 452:
```suggestion
	Context("Pipeline is queued when resources are missing", Ordered, func() {
```

test/e2e/e2e_test.go
```diff
@@ -296,8 +387,193 @@ var _ = Describe("Manager", Ordered, func() {
...
+				cond := apimeta.FindStatusCondition(wl.Status.Conditions, kueue.WorkloadQuotaReserved)
+				if cond == nil {
+					return fmt.Errorf("Didn't find QuotaReserved condition for workload %s", wl.Name)
+				}
+
+				if cond.Status != metav1.ConditionFalse {
+					return fmt.Errorf("QuotaReserved Condition status isn't false (Quota Was reserved while it should have failed)")
+				}
+
+				if !strings.Contains(cond.Message, "insufficient quota for memory") {
```
filariow, line 502: can we use the exact message here?

test/e2e/e2e_test.go
```diff
@@ -296,8 +387,193 @@ var _ = Describe("Manager", Ordered, func() {
...
+	k8sCache, err := cache.New(cfg, cache.Options{Scheme: scheme})
+	Expect(err).ToNot(HaveOccurred(), "failed to create cache")
+
+	err = jobframework.SetupWorkloadOwnerIndex(
+		ctx,
+		k8sCache,
+		tekv1.SchemeGroupVersion.WithKind("PipelineRun"),
+	)
+
+	Expect(err).ToNot(HaveOccurred(), "failed to setup indexer")
```
filariow, line 556: If you like we can simplify it as 

```suggestion
	Expect(jobframework.SetupWorkloadOwnerIndex(
		ctx,
		k8sCache,
		tekv1.SchemeGroupVersion.WithKind("PipelineRun"),
	)).To(Succeed(), "failed to setup indexer")
```

test/e2e/e2e_test.go
```diff
@@ -296,8 +387,193 @@ var _ = Describe("Manager", Ordered, func() {
...
+			Scheme: scheme,
+		},
+	)
+	Expect(err).ToNot(HaveOccurred(), "failed to create client")
+
+	go func() {
+		err := k8sCache.Start(ctx)
+		if err != nil {
+			panic(err)
+		}
```
filariow, line 571: Should we wait for cache to synch?

```suggestion
		if err := k8sCache.Start(ctx); err != nil {
			panic(err)
		}

		if synced := k8sCache.WaitForCacheSync(ctx); !synced {
			panic("failed waiting for cache to sync")
		}
```

test/utils/utils.go
```diff
@@ -33,7 +33,6 @@ const (
 		"releases/download/%s/bundle.yaml"
 
 	certmanagerVersion = "v1.16.3"
```
filariow, line 35: we have this here and in the Makefile. I'd propose to have it in the Makefile only.

Makefile
```diff
@@ -223,3 +228,26 @@ mv $(1) $(1)-$(3) ;\
...
+	kubectl apply --server-side -f https://github.com/cert-manager/cert-manager/releases/download/$(CERT_MANAGER_VERSION)/cert-manager.yaml
+	kubectl wait --for=condition=Available deployment --all -n cert-manager --timeout=300s
+
+.PHONY: cert-manager-undeploy
+cert-manager-undeploy:
+	kubectl delete -f https://github.com/cert-manager/cert-manager/releases/download/$(CERT_MANAGER_VERSION)/cert-manager.yaml
+
+.PHONY: load-image
+load-image: docker-build
+	dir=$$(mktemp -d) && podman save $(IMG) -o $${dir}/tekton-kueue.tar && kind load image-archive -n $(KIND_CLUSTER)  $${dir}/tekton-kueue.tar
```
filariow, line 253: if everything goes fine, shouldn't we delete the folder too?

```suggestion
	dir=$$(mktemp -d) && podman save $(IMG) -o $${dir}/tekton-kueue.tar && kind load image-archive -n $(KIND_CLUSTER)  $${dir}/tekton-kueue.tar && rm -r "${{dir}}"
```

README.md
```diff
@@ -1,135 +1,132 @@
...
+**Install Kueue:**
 
-**Install the CRDs into the cluster:**
+The controller currently supports kueue v0.10.x
+If you already have kueue installed, make sure to enable it by adding `pipelineruns.tekton.dev` to the external frameworks.
 
 ```sh
-make install
+make kueue
+```
```
filariow, line 34: will this `make kueue` "enable it by adding `pipelineruns.tekton.dev` to the external frameworks."? or in that case the user should NOT run `make kueue`?

gbenhaim: when running `make kueue` we deploy the config in `hack/kueue-config.yaml` which has `pipelineruns.tekton.dev` in the external framework.

test/e2e/e2e_test.go
gbenhaim: I found it useful to keep them for debugging purposes.

test/e2e/e2e_test.go
gbenhaim: I  don't really like to really on assertion like that, since there is no guarantee  that the message won't change in newer version of Kueue (changing the message isn't a breaking api change), thus I prefer to limit the text I'm looking in the message.

.github/workflows/test-e2e.yml
```diff
@@ -29,8 +27,19 @@ jobs:
...
+          IMG: konflux-ci.dev/tekton-kueue:v0.0.1
         run: |
           go mod tidy
-          make test-e2e
+          make load-image "IMG=$IMG"
+          podman exec kind-control-plane crictl images
+          make kueue
+          make tekton
+          ret=0
+          make test-e2e || ret=$?
```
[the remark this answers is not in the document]
gbenhaim: removed

test/e2e/e2e_test.go
gbenhaim: removed

test/utils/utils.go
gbenhaim: done

README.md
filariow: Ok. I find the sentence above a bit misleading then. Maybe we can rephrase like follows:

````md
The controller currently supports kueue v0.10.x.
If you already have kueue installed, make sure to enable it by adding `pipelineruns.tekton.dev` to the external frameworks.
Otherwise you can install it with:

```sh
make kueue
```

````

gbenhaim: done

Makefile
gbenhaim: done

Issue #29, opened 2025-05-07
Support pipeline count


gbenhaim:
In the function which generate the PodSpec, add a resource for tracking the number of pipeline runs.
The name of the resource should be `tekton.dev/pipelineruns` and its value should be hardcoded to 1.
By doing that, we can allow users to limit the number of running pipelines.

Issue #28, opened 2025-05-05
Support resource name with domain


gbenhaim:
It's possible to extend k8s with custom resource request and limits (in addition to the built in cpu/memory). It's advised by the k8s documentation to use a domain name in custom resources, for example nvidia.com/gpu.
The annotation for declaring resource requests for PipelineRuns should support custom resources.

PR #33, opened 2025-05-11
Support Pipelinerun count

gbenhaim:
1. Add special resource to the Workload created for a PipelinRun. The resource will be used for controlling the amount of concurrent PipelineRuns that can be executed at the same time.

2. Remove the default 1GB memory request, since it's not needed after the addition of the new resource request.

3. Tests refactoring for sharing assertion and making them more informative.

closes: https://github.com/konflux-ci/tekton-kueue/issues/29

### DIFF
--- a/internal/controller/pipelinerun_controller.go
+++ b/internal/controller/pipelinerun_controller.go
@@ -43,8 +43,8 @@ const (
 )
 
 const (
-	ControllerName              = "KueuePipelineRunController"
-	ResourcePipelineRunCount    = "tekton.dev/pipelineruns"
+	ControllerName           = "KueuePipelineRunController"
+	ResourcePipelineRunCount = "tekton.dev/pipelineruns"
 )
 
 const (
--- a/internal/controller/pipelinerun_controller.go
+++ b/internal/controller/pipelinerun_controller.go
@@ -44,7 +44,7 @@ const (
 
 const (
 	ControllerName              = "KueuePipelineRunController"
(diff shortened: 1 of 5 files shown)


internal/controller/pipelinerun_controller.go
```diff
@@ -179,14 +179,15 @@ func (p *PipelineRun) PodSets() []kueue.PodSet {
...
+// By default, a resource which indicates that the workload requires 1
+// PipelineRun will be added. This is useful for controlling the number
+// of PipelineRuns that can be executed concurrently.
 //
 // WARNING: Annotations are not validated and a panic will
 // happen if they can not be parsed as `resource.Quantity`.
 func (p *PipelineRun) resourcesRequests() corev1.ResourceList {
 	requests := corev1.ResourceList{
-		corev1.ResourceMemory: resource.MustParse(defaultMemoryRequest),
+		PipelineRunCountResourceKey: resource.MustParse("1"),
```
filariow, line 190: I'd suggest ahering to corev1's standard
```suggestion
		ResourcePipelineRunCount: resource.MustParse("1"),
```

filariow, line 190: How can we override this?

gbenhaim: once https://github.com/konflux-ci/tekton-kueue/issues/28 is done users would be able to override this resource request using an annotation, even though I don't see a reason why would anyone do it.

filariow (approved):
/lgtm

PR #41, opened 2025-06-08
Separate controller and webhook

gbenhaim:
For the controller we would need leader election.
  
  We've seen the during etcd defrag, controllers which relies on leases
  can restart.
  
  This cause a problem when the webhook server runs in the same process
  with the controller, since the webhook must be always available otherwise
  pipelinerun creation will fail.
  
  In addition, we would probably want to run different number of replicas
  for the webhook (3) and the controller (2).
  
  We are still going to build a single binary and a single container
  image. The binary will have two sub commands for the controller and
  webhook. This approach will optimize the build process.
  
  The deployment manifest were adjusted and cleaned.
  A dedicate service account with less privileges is used for the
  webhook deployment.
  
  e2e tested were adjusted as well to test the deployment of the
  webhook server. Tests which checks the actual functionality weren't
  changed in order to ensure this change doesn't introduce regressions.
  
  closes https://github.com/konflux-ci/tekton-kueue/issues/45


test/e2e/e2e_test.go
```diff
@@ -237,99 +265,109 @@ var _ = Describe("Manager", Ordered, func() {
...
+							"{{ \"\\n\" }}{{ end }}{{ end }}",
+						"-n", namespace,
+					)
+
+					podOutput, err := utils.Run(cmd)
+					g.Expect(err).NotTo(HaveOccurred(), "Failed to retrieve pod information")
+					podNames := utils.GetNonEmptyLines(podOutput)
+					g.Expect(podNames).To(HaveLen(1), "expected 1 pod running")
+					podName := (podNames[0])
+					g.Expect(podName).To(ContainSubstring(nameSubstring))
```
filariow, line 288:
```suggestion
					g.Expect(podNames[0])To(ContainSubstring(nameSubstring))
```
or
```suggestion
					podName := podNames[0]
					g.Expect(podName).To(ContainSubstring(nameSubstring))
```

test/e2e/e2e_test.go
```diff
@@ -237,99 +265,109 @@ var _ = Describe("Manager", Ordered, func() {
...
-				"--image=curlimages/curl:latest",
-				"--overrides",
-				fmt.Sprintf(`{
+				crb, err := utils.Run(cmd)
+				Expect(err).NotTo(HaveOccurred(), "Failed to generate ClusterRoleBinding")
+
+				cmd = exec.Command("kubectl", "apply", "-f", "-")
+				cmd.Stdin = strings.NewReader(crb)
+				_, err = utils.Run(cmd)
+				Expect(err).NotTo(HaveOccurred(), "Failed to apply ClusterRoleBinding")
```
filariow, line 329:
```suggestion
				Expect(utils.Run(cmd)).To(Succeed(), "Failed to apply ClusterRoleBinding")
```

test/e2e/e2e_test.go
```diff
@@ -237,99 +265,109 @@ var _ = Describe("Manager", Ordered, func() {
...
+
+				By("validating that the metrics service is available")
+				cmd = exec.Command("kubectl", "get", "service", metricsServiceName, "-n", namespace)
+				_, err = utils.Run(cmd)
+				Expect(err).NotTo(HaveOccurred(), "Metrics service should exist")
```
filariow, line 334:
```suggestion
				Expect(utils.Run(cmd)).To(Succeed(), "Metrics service should exist")
```

test/e2e/e2e_test.go
```diff
@@ -237,99 +265,109 @@ var _ = Describe("Manager", Ordered, func() {
...
+
+				By("validating that the ServiceMonitor for Prometheus is applied in the namespace")
+				cmd = exec.Command("kubectl", "get", "ServiceMonitor", "-n", namespace)
+				_, err = utils.Run(cmd)
+				Expect(err).NotTo(HaveOccurred(), "ServiceMonitor should exist")
```
filariow, line 339:
```suggestion
				Expect(utils.Run(cmd)).To(Succeed(), "ServiceMonitor should exist")
```

test/e2e/e2e_test.go
```diff
@@ -351,26 +389,39 @@ var _ = Describe("Manager", Ordered, func() {
...
-				g.Expect(err).NotTo(HaveOccurred())
-				g.Expect(output).To(Equal("Succeeded"), "curl pod in wrong status")
-			}
-			Eventually(verifyCurlUp, 5*time.Minute).Should(Succeed())
-
-			By("getting the metrics by checking curl-metrics logs")
-			metricsOutput := getMetricsOutput()
-			Expect(metricsOutput).To(ContainSubstring(
+				_, err = utils.Run(cmd)
+				Expect(err).NotTo(HaveOccurred(), "Failed to create pod", getPodName())
```
filariow, line 393:
```suggestion
				Expect(utils.Run(cmd)).To(Succeed(), "Failed to create pod", getPodName())
```

test/e2e/e2e_test.go
```diff
@@ -237,99 +265,109 @@ var _ = Describe("Manager", Ordered, func() {
...
+					// Validate the pod's status
+					cmd = exec.Command("kubectl", "get",
+						"pods", podName, "-o", "jsonpath={.status.phase}",
+						"-n", namespace,
+					)
+					output, err := utils.Run(cmd)
+					g.Expect(err).NotTo(HaveOccurred())
+					g.Expect(output).To(Equal("Running"), "Incorrect pod status")
+
+					podNamerSetter(podName)
```
filariow, line 299: typo

```suggestion
			func(label string, nameSubstring string, podNameSetter PodNameSetter) {
				By("validating that the controller-manager pod is running as expected")
				verifyControllerUp := func(g Gomega) {
					// Get the name of the controller-manager pod
					cmd := exec.Command("kubectl", "get",
						"pods", "-l", label,
						"-o", "go-template={{ range .items }}"+
							"{{ if not .metadata.deletionTimestamp }}"+
							"{{ .metadata.name }}"+
							"{{ \"\\n\" }}{{ end }}{{ end }}",
						"-n", namespace,
					)

					podOutput, err := utils.Run(cmd)
					g.Expect(err).NotTo(HaveOccurred(), "Failed to retrieve pod information")
					podNames := utils.GetNonEmptyLines(podOutput)
					g.Expect(podNames).To(HaveLen(1), "expected 1 pod running")
					podName := (podNames[0])
					g.Expect(podName).To(ContainSubstring(nameSubstring))

					// Validate the pod's status
					cmd = exec.Command("kubectl", "get",
						"pods", podName, "-o", "jsonpath={.status.phase}",
						"-n", namespace,
					)
					output, err := utils.Run(cmd)
					g.Expect(err).NotTo(HaveOccurred())
					g.Expect(output).To(Equal("Running"), "Incorrect pod status")

					podNameSetter(podName)
```

filariow (changes requested):

test/e2e/e2e_test.go
gbenhaim: I don't think this is possible since utils.run has 2 return values.

test/e2e/e2e_test.go
gbenhaim: I don't think this is possible since utils.run has 2 return values.

config/webhook/kustomization.yaml
```diff
@@ -1,6 +1,13 @@
+kind: Kustomization
+images:
+- name: controller
+  newName: konflux-ci/tekton-kueue
+  newTag: latest
```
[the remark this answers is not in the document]
gbenhaim: it's needed when deploying using the Makefile (used as part of the tests)

config/default/kustomization.yaml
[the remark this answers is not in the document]
gbenhaim: good catch. done.

test/e2e/e2e_test.go
gbenhaim: done

test/e2e/e2e_test.go
gbenhaim: done

PR #47, opened 2025-06-10
Make the default queue name configurable

pramodbindal:


cmd/main.go
```diff
@@ -221,7 +221,7 @@ func main() {
 
 	err = webhookv1.SetupPipelineRunWebhookWithManager(
 		mgr,
-		&webhookv1.PipelineRunCustomDefaulter{KueueName: "pipelines-queue"},
+		&webhookv1.PipelineRunCustomDefaulter{Client: mgr.GetClient()},
```
gbenhaim, line 224: I think we should mount the configmap to the controller. this way when it changes it will restart and the new values will get reflected.

gbenhaim (changes requested):

cmd/main.go
pramodbindal: configMap is created in tenant's namespace so every  tenant may have different value for default-queue name.


mounting the configMap might not be feasible here but yes instead of reading the value on each request we can think of caching the value and update the cache when configMap changes

pramodbindal: Done

config/rbac/role.yaml
```diff
@@ -72,9 +72,3 @@ rules:
   - patch
   - update
   - watch
-- apiGroups:
```
gbenhaim, line 75: I think this file was committed by mistake. we need to keep those permissions.

cmd/main.go
```diff
@@ -216,10 +223,15 @@ func runWebhook(args []string) {
...
+	cfg, err := loadConfig(webhookFlags.ConfigDir)
+	if err != nil {
+		setupLog.Error(err, "unable to load webhook configuration")
+		os.Exit(1)
+	}
 
 	err = webhookv1.SetupPipelineRunWebhookWithManager(
 		mgr,
-		&webhookv1.PipelineRunCustomDefaulter{KueueName: "pipelines-queue"},
+		&webhookv1.PipelineRunCustomDefaulter{Config: cfg},
```
gbenhaim, line 234: I think it would be better to just pass `cfg.QueueName` instead of the entire config in order to not expose `PipelineRunCustomDefaulter` to information it doesn't need, or in other words, control exactly which information it can access.

internal/config/config_type.go
```diff
@@ -0,0 +1,25 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+const (
+	DefaultQueueName = "pipelines-queue"
```
gbenhaim, line 20: after looking at this default I have second thoughts.
WDYT about making the configmap required and include a configmap with defaults in our manifests?
This way we don't branch our code (loading a config is always required) and this path will be tested as part of the e2e tests we already have.

gbenhaim (changes requested):

internal/config/config_type.go
pramodbindal: we can do that. anyways we are creating configmap as part of deploy  steps

internal/webhook/v1/pipelinerun_webhook.go
```diff
@@ -23,17 +23,18 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
-// nolint:unused
-// log is for logging in this package.
-var pipelinerunlog = logf.Log.WithName("pipelinerun-resource")
-
 // SetupPipelineRunWebhookWithManager registers the webhook for PipelineRun in the manager.
 func SetupPipelineRunWebhookWithManager(mgr ctrl.Manager, defaulter admission.CustomDefaulter) error {
+	d := defaulter.(*PipelineRunCustomDefaulter) // Type assert to PipelineRunCustomDefaulter
```
gbenhaim, line 32: Instead of breaking the abstraction and requiring to call the Validate function, I suggest to:

1. Make the PipelineRunsCustomDefaulter private
2. Create a function called NewPipelineRunsCustomDefaulter (with return type admission.CustomDefaulter) and do the validation in this function.

With this approach we won't break the abstrection and it won't be possible to create an instance with the wrong configurations.

gbenhaim (changes requested):
Added one last comment about the Validate function you added.

Thanks for working on this !

PR #50, opened 2025-06-11
feat: Add workflow to build container images with ko

chmouel:
- Added a new GitHub Actions workflow (`container-build.yaml`).
- Configured the workflow to use `ko-build` for container image creation.
- Enabled building images for `linux/amd64`, `linux/arm64`, and `linux/ppc64le` platforms.
- Set up automatic push of built images to the GitHub Container Registry (GHCR).
- Triggered the workflow on push events involving Go or YAML files.

### DIFF
--- a/.github/workflows/container-build.yaml
+++ b/.github/workflows/container-build.yaml
@@ -1,0 +1,48 @@
+name: Create and publish a Docker image to ghcr on main and nightly with ko
+
+on:
+  push:
+    paths:
+      - "**.go"
+      - "**.yaml"
+env:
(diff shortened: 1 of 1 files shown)


gbenhaim:
@chmouel would it be maybe better to build for arm using Konflux? we already build the x86 image using Konflux.

pramodbindal:
As we are building this repo in konflux via Dockerfile. We should utilize the same Dockerfile for all build purposes.
We have seen in past the behavioral differences in the images built from Dockerfile and KO

PR #51, opened 2025-06-12
Upgrade Kueue to 0.11.6

pramodbindal:
This PR is to upgrade Kueue package to 0.11.6


pramodbindal:
@gbenhaim PTAL

test/utils/utils.go
```diff
@@ -180,7 +180,7 @@ func GetProjectDir() (string, error) {
 	if err != nil {
 		return wd, err
 	}
-	wd = strings.Replace(wd, "/test/e2e", "", -1)
+	wd = strings.ReplaceAll(wd, "/test/e2e", "")
```
gbenhaim, line 183: why this change is needed?

internal/controller/pipelinerun_controller.go
```diff
@@ -207,3 +207,5 @@ func (p *PipelineRun) RunWithPodSetsInfo(podSetsInfo []podset.PodSetInfo) error
 func (p *PipelineRun) Suspend() {
 	// Not implemented because this is not called when JobWithCustomStop is implemented.
 }
+
+// Suspend implements jobframework.GenericJob.
```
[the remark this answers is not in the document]
pramodbindal: Done

test/utils/utils.go
chmouel: golangci-lint should warn to prefer using ReplaceAll as well, it may be a good idea to enforce it in CI

PR #80, opened 2025-07-15
Add Build and Release Workflow

pramodbindal:


.github/workflows/build-and-release-tekton-kueue.yaml
```diff
@@ -0,0 +1,13 @@
...
+on:
+  workflow_dispatch: { }
+  push:
+    branches:
+      - main
+      - release-*
+
+jobs:
+  release:
+    uses: pramodbindal/hack/.github/workflows/build-and-release.yaml@main
```
gbenhaim: Is this going to create a github release?

hack/build.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# 👇 Export to GitHub Actions environment if running in CI
+if [ -n "$GITHUB_ENV" ]; then
+  echo "VERSION=$VERSION" >> "$GITHUB_ENV"
+fi
+
+export IMG=$DOCKER_REPO:$VERSION
+echo $IMG
+
+#make manifests kustomize  docker-buildx
```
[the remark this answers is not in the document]
pramodbindal: Yes that makes better sense. How can we get the  $IMG and $TAG from konflux ?
Can we something in release pipeline which updates the latest SHA somewhere in repo

config/manager/kustomization.yaml
```diff
@@ -5,5 +5,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: konflux-ci/tekton-kueue
+  newName: quay.io/openshift-pipeline/tekton-kueue
```
gbenhaim: This will require a change to the overlays in infra deployments after it get merged. For example - https://github.com/redhat-appstudio/infra-deployments/blob/461ec1385fbaab3e84a6f35454e1fffd438028a2/components/kueue/development/tekton-kueue/kustomization.yaml#L7

The name of the new image would need to be updated in each overlay: development, staging and production.

pramodbindal: no issues we can keep it as it is

gbenhaim (approved):

Issue #90, opened 2025-08-18
Support summing resource requests for duplicate keys when using CEL


gbenhaim:
In order to make it easier for users to write CEL expressions that generate resource request, we should support summation of duplicate keys.
For example,

```cel
[
   resource("aws-vm-x", 2),
   resource("aws-vm-x", 1)
]
```

should set the following annotation

`kueue.konflux-ci.dev/requests-aws-vm-x: "3"`

PR #91, opened 2025-08-18
feat: Add resource annotation summation for duplicate keys

gbenhaim:
Add ability to automatically sum resource request values when multiple CEL expressions target the same resource key. Previously, duplicate resource requests would overwrite each other - now they aggregate.

This solves the problem where complex pipelines with multiple steps or conditional branches need to accumulate resource requirements:

- Multiple pipeline tasks requesting the same VM type now sum quantities
- Conditional expressions can safely add resources without conflicts
- Eliminates need for manual resource calculation in complex workflows
- Maintains type safety with positive integer validation

Example: resource("aws-vm-x", 2) + resource("aws-vm-x", 3) = 5 total
requests, automatically creating kueue.konflux-ci.dev/requests-aws-vm-x: "5"

This enables more sophisticated resource planning for multi-step pipelines without resource request conflicts.

closes https://github.com/konflux-ci/tekton-kueue/issues/90

Assisted-By: Cursor

### DIFF
--- a/internal/cel/compiler.go
+++ b/internal/cel/compiler.go
@@ -56,6 +56,7 @@ func createCELEnvironment() (*cel.Env, error) {
 		// Add type-safe functions for creating MutationRequests
 		createMutationFunction("annotation", MutationTypeAnnotation, mutationRequestType),
 		createMutationFunction("label", MutationTypeLabel, mutationRequestType),
+		createResourceMutationFunction("resource", MutationTypeResource, mutationRequestType),
 		createPriorityMutationFunction("priority", mutationRequestType),
 		// Add string manipulation functions
 		createReplaceFunction("replace"),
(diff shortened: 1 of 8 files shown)


internal/cel/mutator.go
```diff
@@ -98,6 +103,28 @@ func mutate(pipelineRun *tekv1.PipelineRun, mutation *MutationRequest) *tekv1.Pi
...
+		newValue, err := strconv.Atoi(mutation.Value)
+		if err != nil {
+			// If parsing fails, treat as 0
+			newValue = 0
+		}
+
+		// Check if the key already exists and sum the values
+		if existingValue, exists := pipelineRun.Annotations[mutation.Key]; exists {
+			if existingInt, err := strconv.Atoi(existingValue); err == nil {
+				newValue += existingInt
```
filariow, line 121: ditto

filariow (changes requested):
mostly lgtm, some minor feedback and a concern on error suppression in `cel/mutator`

internal/cel/types.go
```diff
@@ -12,12 +12,13 @@ type MutationType string
...
 	MutationTypeAnnotation MutationType = "annotation"
 	MutationTypeLabel      MutationType = "label"
+	MutationTypeResource   MutationType = "resource"
 )
 
 // IsValid checks if the mutation type is valid
 func (mt MutationType) IsValid() bool {
 	switch mt {
-	case MutationTypeAnnotation, MutationTypeLabel:
+	case MutationTypeAnnotation, MutationTypeLabel, MutationTypeResource:
```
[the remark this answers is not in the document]
gbenhaim: done

internal/cel/mutator.go
gbenhaim: done

gbenhaim:
@filariow thanks for the revie. I made the fixes. Anything else? Can I merge it?

internal/cel/compiler.go
```diff
@@ -129,6 +130,65 @@ func createMutationFunction(name string, mutationType MutationType, returnType *
...
+
+				// Convert int value to string for storage
+				value := fmt.Sprintf("%d", intValue)
+
+				// Validate key using annotation validation since resource mutations create annotations
+				err := validateKey(key, "resource annotation")
+				if err != nil {
+					return types.NewErr("%s key validation failed: %v", name, err)
+				}
+
```
filariow, line 171: minor: no need to convert value to string if key is invalid

```suggestion
				// Validate key using annotation validation since resource mutations create annotations
				err := validateKey(key, "resource annotation")
				if err != nil {
					return types.NewErr("%s key validation failed: %v", name, err)
				}

				// Convert int value to string for storage
				value := fmt.Sprintf("%d", intValue)

```

internal/cel/mutator.go
```diff
@@ -98,6 +110,30 @@ func mutate(pipelineRun *tekv1.PipelineRun, mutation *MutationRequest) *tekv1.Pi
...
+	case MutationTypeResource:
+		if pipelineRun.Annotations == nil {
+			pipelineRun.Annotations = make(map[string]string)
+		}
+
+		// Parse the new value as integer
+		newValue, err := strconv.Atoi(mutation.Value)
+		if err != nil {
+			// This should never happen because we validate the value in the CEL compiler
+			return nil, fmt.Errorf("failed to parse resource value %q as integer: %w", mutation.Value, err)
```
filariow, line 122: if I understand it correctly, this means there is a bug somewhere. should we panic here instead?

internal/cel/mutator.go
```diff
@@ -98,6 +110,30 @@ func mutate(pipelineRun *tekv1.PipelineRun, mutation *MutationRequest) *tekv1.Pi
 			pipelineRun.Annotations = make(map[string]string)
 		}
 		pipelineRun.Annotations[mutation.Key] = mutation.Value
+	case MutationTypeResource:
+		if pipelineRun.Annotations == nil {
+			pipelineRun.Annotations = make(map[string]string)
+		}
+
+		// Parse the new value as integer
+		newValue, err := strconv.Atoi(mutation.Value)
```
filariow, line 119: in `createResourceMutationFunction` we are using `intValue, intValueOk := rhs.Value().(int64)`, I think we should prefer `strconv.ParseInt` to ensure int64 is used instead of `strconv.Atoi`.

```suggestion
		newValue, err := strconv.ParseInt(mutation.Value, 10, 64)
```

internal/cel/mutator.go
```diff
@@ -98,6 +110,30 @@ func mutate(pipelineRun *tekv1.PipelineRun, mutation *MutationRequest) *tekv1.Pi
...
+		if err != nil {
+			// This should never happen because we validate the value in the CEL compiler
+			return nil, fmt.Errorf("failed to parse resource value %q as integer: %w", mutation.Value, err)
+		}
+
+		// Check if the key already exists and sum the values
+		if existingValue, exists := pipelineRun.Annotations[mutation.Key]; exists {
+			existingInt, err := strconv.Atoi(existingValue)
```
filariow, line 127: ditto

```suggestion
			existingInt, err := strconv.ParseInt(mutation.Value, 10, 64)
```

filariow (approved):
/lgtm

some minor feedback. We can defer them to a follow-up PR if needed.

internal/cel/mutator.go
gbenhaim: I prefer to log the issue and report it using metrics instead of panic so other requests can be still served.

gbenhaim:
I'll create a follow-up pr next week when I'll be back.

PR #175, opened 2025-12-11
Update Kueue to v0.14.5

gbenhaim:
Required some changes to the code because of breaking API changes, nothing which changes functionality.


Dockerfile
```diff
@@ -1,5 +1,6 @@
 # Build the manager binary
-FROM registry.access.redhat.com/ubi9/go-toolset@sha256:7b1828de52c3bac600a71b81996bf748776a456181a45e2b329b39702cf6486f AS builder
+# golang 1.25.3
+FROM registry.access.redhat.com/ubi9/go-toolset@sha256:56fc8480721e5febbd2b6810199a63b5b68c3c5ab826d082c8247054a1dc33e7 AS builder
```
pramodbindal: can we use tag instead ?
using label will help us identify the golang version

PR #247, opened 2025-12-23
Add Tekton-Kueue-Config with watcher

pramodbindal:
Currently Config is loaded at pod startup time and if there is any change in config then pod must be restarted  for new changes to take effect.

With updated config, if there is any change in config then reconciler loads that immediately and webhook will start using the updated config.

Config map structure is updated to accomodate the multi-cluster feature

```
---
apiVersion: v1
data:
  config.yaml: |
    queueName: pipelines-queue
    cel:
      expressions:
        - priority("tekton-kueue-default")
kind: ConfigMap
metadata:
  labels:
    app.kubernetes.io/part-of: tekton-kueue
  name: tekton-kueue-config
  namespace: tekton-kueue
```

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -19,7 +19,6 @@ package main
 import (
 	"context"
 	"crypto/tls"
-	"errors"
 	"flag"
 	"fmt"
 	"os"
@@ -28,7 +27,9 @@ import (
 	"reflect"
 	"time"
 
-	"k8s.io/apimachinery/pkg/util/yaml"
+	"github.com/konflux-ci/tekton-kueue/internal/common"
+	"gopkg.in/yaml.v3"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 	outputyaml "sigs.k8s.io/yaml"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
@@ -48,10 +49,8 @@ import (
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
-	"github.com/konflux-ci/tekton-queue/internal/cel"
-	kueueconfig "github.com/konflux-ci/tekton-queue/internal/config"
-	"github.com/konflux-ci/tekton-queue/internal/controller"
-	webhookv1 "github.com/konflux-ci/tekton-queue/internal/webhook/v1"
+	"github.com/konflux-ci/tekton-kueue/internal/controller"
+	webhookv1 "github.com/konflux-ci/tekton-kueue/internal/webhook/v1"
 
 	// +kubebuilder:scaffold:imports
 
@@ -259,6 +258,7 @@ func runController(args []string) {
 }
 
 func runWebhook(args []string) {
+	setupLog.Info("starting webhook")
 	fs := flag.NewFlagSet("webhook", flag.ExitOnError)
 	var webhookFlags WebhookFlags
 	webhookFlags.AddFlags(fs)
@@ -269,35 +269,31 @@ func runWebhook(args []string) {
 
 	webhookOptions, webhookCertWatcher := getWebhookServerOptions(webhookFlags, tlsOpts)
 	webhookServer := webhook.NewServer(webhookOptions)
-
+	namespace, err := common.GetCurrentNamespace()
+	if err != nil {
+		setupLog.Error(err, "Exiting")
+		os.Exit(1)
+	}
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:                 scheme,
 		Metrics:                metricsServerOptions,
 		HealthProbeBindAddress: webhookFlags.ProbeAddr,
 		WebhookServer:          webhookServer,
 		LeaderElection:         false,
+		Cache: cache.Options{
+			DefaultNamespaces: map[string]cache.Config{
+				namespace: {}, // namespace where SA has access
+			},
+		},
 	})
 	if err != nil {
 		setupLog.Error(err, "unable to create manager")
 		os.Exit(1)
 	}
-	cfg, err := loadConfig(webhookFlags.ConfigDir)
+	cfgStore := &webhookv1.ConfigStore{}
+	customDefaulter, err := webhookv1.NewCustomDefaulter(cfgStore)
 	if err != nil {
-		setupLog.Error(err, "unable to load webhook configuration")
-		os.Exit(1)
-	}
-
-	programs, err := cel.CompileCELPrograms(cfg.CEL.Expressions)
-	if err != nil {
-		setupLog.Error(err, "unable to compile CEL programs")
-		os.Exit(1)
-	}
-	mutator := cel.NewCELMutator(programs)
-
-	customDefaulter, err := webhookv1.NewCustomDefaulter(cfg, []webhookv1.PipelineRunMutator{mutator})
-
-	if err != nil {
-		setupLog.Error(err, "Unable to create custom defaulter for webhook")
+		setupLog.Error(err, "unable to create custom defaulter")
 		os.Exit(1)
 	}
 	err = webhookv1.SetupPipelineRunWebhookWithManager(
@@ -308,6 +304,7 @@ func runWebhook(args []string) {
 		setupLog.Error(err, "Failed to setup the webhook")
 		os.Exit(1)
 	}
+	addConfigWatcher(mgr, cfgStore)
 	addRunnableOrDie(
 		mgr,
 		webhookCertWatcher,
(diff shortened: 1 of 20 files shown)


gbenhaim:
I still need to go over the entire pr, first thing that popes, why do you call the config "scheduler-config"? tekton-kueue doesn't have any scheduler in it.

cmd/main.go
```diff
@@ -542,28 +509,54 @@ func addMetricsCertWatcher(mgr ctrl.Manager, runnable manager.Runnable) {
...
 
+func addConfigWatcher(mgr ctrl.Manager, configStore *webhookv1.ConfigStore) {
+	setupLog.Info("Adding config watcher to manager")
+	namespace, found := os.LookupEnv("POD_NAMESPACE")
+	if !found {
+		setupLog.Error(fmt.Errorf("POD_NAMESPACE env var not set"), "Exiting")
+		os.Exit(1)
+		// ns will be empty string "", which in K8s terms means "all namespaces"
+	}
+	err := ctrl.NewControllerManagedBy(mgr).
```
gbenhaim: the convention is to add SetupWithManager method to the reconciler and place it in the same file with the reconciler.

```
func (r *ConfigmapReconciler) SetupWithManager(mgr ctrl.Manager) error {
    return ctrl.NewControllerManagedBy(mgr).
                 Named("webhook-config").
		For(&corev1.ConfigMap{}, predicate.NewPredicateFuncs(func(o client.Object) bool {
			return o.GetName() == common.ConfigMapName && o.GetNamespace() == namespace
		})).
		Complete(r)

}

File: config/webhook/deployment.yaml [hunk 2]
```diff
@@ -27,61 +27,66 @@ spec:
...
-            cpu: 10m
-            memory: 64Mi
-        volumeMounts:
-        - mountPath: /tmp/k8s-webhook-server/serving-certs
-          name: webhook-certs
-          readOnly: true
-        - mountPath: /tmp/k8s-webhook-server/kueue-config
-          name: kueue-config
-          readOnly: true
+        - command:
```
gbenhaim: a lot of line here changed without a reason. probably because of indentation. can you please change only the required lines so it would be easy to review?

internal/webhook/v1/pipelinerun_webhook.go
```diff
@@ -124,24 +114,18 @@ func (d *pipelineRunCustomDefaulter) Default(ctx context.Context, obj runtime.Ob
...
 		plr.Labels = make(map[string]string)
 	}
-	if _, exists := plr.Labels[common.QueueLabel]; !exists {
-		plr.Labels[common.QueueLabel] = d.config.QueueName
+	config := d.configStore.GetConfig()
+	if label, exists := plr.Labels[common.QueueLabel]; !exists || label == "" {
+		plr.Labels[common.QueueLabel] = config.QueueName
 	}
-	if d.config.MultiKueueOverride {
+	if config.MultiClusterEnabled && config.MultiClusterRole == common.MultiClusterRoleHub {
```
gbenhaim: why do we need to introduce the concepts of roles? what is wrong with MultiKueueOverride ?

internal/webhook/v1/pipelinerun_webhook.go
```diff
@@ -124,24 +114,18 @@ func (d *pipelineRunCustomDefaulter) Default(ctx context.Context, obj runtime.Ob
 	if plr.Labels == nil {
 		plr.Labels = make(map[string]string)
 	}
-	if _, exists := plr.Labels[common.QueueLabel]; !exists {
-		plr.Labels[common.QueueLabel] = d.config.QueueName
+	config := d.configStore.GetConfig()
+	if label, exists := plr.Labels[common.QueueLabel]; !exists || label == "" {
```
gbenhaim: this change is unrelated. I also don't understand why it's required. if the user and empty label we should respect it.

gbenhaim (changes requested):
Thanks for working on this. I've posted some comments.

config/rbac/webhook_role.yaml
```diff
@@ -0,0 +1,13 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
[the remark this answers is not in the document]
pramodbindal: I tried with that.
For that we need to limit the scope of watcher for specific namespace.

but same mgr is watching the pipelineruns from all namespaces

go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/konflux-ci/tekton-queue
+module github.com/konflux-ci/tekton-kueue
```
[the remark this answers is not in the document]
pramodbindal: repo name and  go module name must match otherwise we cannot import this in operator

internal/webhook/v1/pipelinerun_webhook.go
pramodbindal: In a multicluster environment a cluster can be Hub or spoke.


Will  split the PR in two parts.

internal/webhook/v1/pipelinerun_webhook.go
pramodbindal: yes this is unintended. will revert

internal/webhook/v1/config_store.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func (s *ConfigStore) GetConfig() config.Config {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.config
+}
+
+func (s *ConfigStore) GetMutators() []PipelineRunMutator {
+	s.mu.Lock()
```
gbenhaim: when doing read operations you should use RLock (read lock) so multiple go routines which servers request can read the config.

internal/webhook/v1/config_store.go
```diff
@@ -0,0 +1,68 @@
...
+func validateConfig(config config.Config) error {
+	if config.QueueName == "" {
+		return errors.New("queue name is not set in the PipelineRunCustomDefaulter")
+	}
+	return nil
+}
+
+func parseConfig(raw []byte) (config.Config, error) {
+	cfg := config.Config{}
+	if err := yaml.Unmarshal(raw, &cfg); err != nil {
```
gbenhaim: we should create a Prometheus metric that shows the number of successful/failed configmap reloading. administrators can use this metric for setting up alert that will fire in case of a bad config.

internal/webhook/v1/config_store.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+type ConfigStore struct {
+	mu       sync.RWMutex
+	config   config.Config
+	mutators []PipelineRunMutator
+}
+
+func (s *ConfigStore) GetConfig() config.Config {
+	s.mu.Lock()
```
gbenhaim: same comment as below about using RLock

cmd/main.go
```diff
@@ -281,33 +283,17 @@ func runWebhook(args []string) {
...
-
-	programs, err := cel.CompileCELPrograms(cfg.CEL.Expressions)
-	if err != nil {
-		setupLog.Error(err, "unable to compile CEL programs")
-		os.Exit(1)
-	}
-	mutator := cel.NewCELMutator(programs)
-
-	customDefaulter, err := webhookv1.NewCustomDefaulter(cfg, []webhookv1.PipelineRunMutator{mutator})
+	cfgStore := &webhookv1.ConfigStore{}
```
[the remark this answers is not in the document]
gbenhaim: why it's needed? the webhook only reference one configstore.

cmd/main.go
```diff
@@ -542,28 +509,54 @@ func addMetricsCertWatcher(mgr ctrl.Manager, runnable manager.Runnable) {
...
-	if dir == "" {
-		return nil, errors.New("no config directory provided")
+func customDefaulter(configDir string, cfgStore *webhookv1.ConfigStore) webhook.CustomDefaulter {
+	setupLog.Info("Loading Kueue config from ", "dir", configDir, "file", "config.yaml")
+	if configDir == "" {
+		setupLog.Error(fmt.Errorf("configDir cannot be empty directory provided"), "Exiting")
+		os.Exit(1)
 	}
-	data, err := os.ReadFile(path.Join(dir, "config.yaml"))
+	data, err := os.ReadFile(path.Join(configDir, "config.yaml"))
```
gbenhaim: why do we need to still mount the config files? I think that having 2 source is confusing. we can wait for the configstore to load the configmap (for the first time) while the controller starts.

internal/webhook/v1/config_store.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"errors"
+	"sync"
+
+	"github.com/konflux-ci/tekton-kueue/internal/cel"
+	"github.com/konflux-ci/tekton-kueue/pkg/config"
+	"gopkg.in/yaml.v3"
+)
+
+type ConfigStore struct {
```
gbenhaim: can you please add some logging to the method in this file so it would be visible in the logs when the config changes?

internal/webhook/v1/config_store.go
```diff
@@ -0,0 +1,68 @@
...
+	if config.QueueName == "" {
+		return errors.New("queue name is not set in the PipelineRunCustomDefaulter")
+	}
+	return nil
+}
+
+func parseConfig(raw []byte) (config.Config, error) {
+	cfg := config.Config{}
+	if err := yaml.Unmarshal(raw, &cfg); err != nil {
+		// Log and keep last-known-good config
```
gbenhaim: Comment says "log" but there's no logging

cmd/main.go
pramodbindal: let me check the delay

cmd/main.go
```diff
@@ -269,35 +269,31 @@ func runWebhook(args []string) {
 
 	webhookOptions, webhookCertWatcher := getWebhookServerOptions(webhookFlags, tlsOpts)
 	webhookServer := webhook.NewServer(webhookOptions)
-
+	namespace, found := os.LookupEnv("POD_NAMESPACE")
```
gbenhaim: you can read this value from /var/run/secrets/kubernetes.io/serviceaccount/namespace
no need to add environment variable and use the downward api for it.

internal/controller/configmap_controller.go
```diff
@@ -0,0 +1,60 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+type ConfigMapReconciler struct {
+	Client client.Client
+	Store  *v1.ConfigStore
+}
+
+func (r *ConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	namespace, found := os.LookupEnv("POD_NAMESPACE")
```
gbenhaim: same comment as above as the namespace. it's probably would be also better to read it once and then pass it as an argument.

gbenhaim (approved):

PR #279, opened 2026-02-09
fix: prevent webhook from overriding default

gbenhaim:
The admission webhook was calling SetDefaults directly on the original PipelineRun object, which permanently set Tekton's built-in 1h timeout on the spec. This overrode cluster-level TektonConfig defaults (e.g. 2h timeout) for PipelineRuns created via CLI without explicit timeouts.

Move the SetDefaults call into CELMutator.Mutate where it operates on a deep copy used only for CEL expression evaluation (structToCELMap needs a fully-defaulted object to marshal successfully). Mutations produced by CEL are applied to the original PipelineRun, leaving its spec untouched.

Remove the inline validation (SetDefaults + Validate + json.Marshal) from the webhook. Instead, introduce a ValidationError type in the cel package that wraps structToCELMap failures. The webhook checks for this error via errors.As and returns k8serrors.NewBadRequest, keeping the validation concern inside the CEL evaluator rather than the webhook.

Assisted-By: Cursor

### DIFF
--- a/internal/webhook/v1/pipelinerun_webhook.go
+++ b/internal/webhook/v1/pipelinerun_webhook.go
@@ -18,10 +18,11 @@ package v1
 
 import (
 	"context"
-	"encoding/json"
+	"errors"
 	"fmt"
 
 	"github.com/go-logr/logr"
+	"github.com/konflux-ci/tekton-kueue/internal/cel"
 	"github.com/konflux-ci/tekton-kueue/pkg/common"
 	tekv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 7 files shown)


internal/webhook/v1/pipelinerun_webhook.go
```diff
@@ -133,6 +112,10 @@ func (d *pipelineRunCustomDefaulter) Default(ctx context.Context, obj runtime.Ob
 	}
 	for _, mutator := range mutators {
 		if err := mutator.Mutate(plr); err != nil {
+			var validationErr *cel.ValidationError
+			if errors.As(err, &validationErr) {
```
filariow: this is a `**cel.ValidationError`, is this right? I was expecting something like the following instead:


```suggestion
			var validationErr cel.ValidationError
			if errors.As(err, &validationErr) {
```

filariow (approved):
mostly lgtm, some minor feedback

PR #325, opened 2026-03-26
fix: use explicit JSON patches to prevent zero-value field leaks in webhook

waveywaves:
# Changes

Replace the `CustomDefaulter` pattern (struct round-trip serialization) with a direct `admission.Handler` that returns explicit JSON patches.

## Problem

The `CustomDefaulter` pattern deserializes the admission request into a typed Go struct, applies mutations, then re-serializes the full struct to compute a diff. This leaks zero-value fields (like `serviceAccountName: ""`) into the patch, preventing the downstream Tekton webhook from applying `default-service-account: pipeline` from `config-defaults`.

Result: PipelineRuns get `serviceAccountName: "default"` instead of `"pipeline"` when the scheduler is enabled.

## Fix

The new `PipelineRunWebhookHandler` implements `admission.Handler` directly and builds JSON patches for only the fields we intend to modify:

- `/spec/status` -> PipelineRunPending
- `/metadata/labels/kueue.x-k8s.io~1queue-name` -> queue name
- `/spec/managedBy` -> multikueue label (if enabled)
- CEL mutations -> labels/annotations only (diffed via `metadataDiffPatches`)

No spec fields beyond `status` and `managedBy` are touched, so downstream webhooks see the object cleanly.

Shared `ApplyMutations()` function ensures the legacy `CustomDefaulter` (used by the mutate CLI) stays in sync with the webhook handler without code duplication.

# Submitter Checklist

- [x] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Release notes block below has been updated with any user facing changes

# Tests

- 5 new unit tests for `PipelineRunWebhookHandler.Handle()`:
  - Sets spec.status and queue label correctly
  - Does not duplicate existing queue label
  - Sets managedBy only when multiKueueOverride is true
  - **Does NOT include serviceAccountName or taskRunTemplate in patches** (core bug fix assertion)
  - Does not set managedBy when multiKueueOverride is false
- 1 new e2e regression test: verifies serviceAccountName is not set to "default" on created PipelineRuns
- All 22 unit tests pass, integration test compiles

# Release Notes

```release-note
Fix webhook mutation leaking zero-value fields (e.g., serviceAccountName) that prevented downstream Tekton webhook from applying default-service-account configuration. The webhook now uses explicit JSON patches instead of struct round-trip serialization.
```

Fixes: https://github.com/konflux-ci/tekton-kueue/issues/319
Related: https://github.com/tektoncd/pipeline/issues/9647

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -45,6 +45,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/metrics/filters"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	"github.com/konflux-ci/tekton-kueue/internal/controller"
 	webhookv1 "github.com/konflux-ci/tekton-kueue/internal/webhook/v1"
@@ -288,14 +289,11 @@ func runWebhook(args []string) {
 		os.Exit(1)
 	}
 	cfgStore := &webhookv1.ConfigStore{}
-	customDefaulter, err := webhookv1.NewCustomDefaulter(cfgStore)
-	if err != nil {
-		setupLog.Error(err, "unable to create custom defaulter")
-		os.Exit(1)
-	}
+	decoder := admission.NewDecoder(scheme)
+	handler := webhookv1.NewWebhookHandler(cfgStore, decoder)
 	err = webhookv1.SetupPipelineRunWebhookWithManager(
 		mgr,
-		customDefaulter,
+		handler,
 	)
 	if err != nil {
 		setupLog.Error(err, "Failed to setup the webhook")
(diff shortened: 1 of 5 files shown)


pramodbindal (approved):

pramodbindal:
@waveywaves Thanks for working on this 

approach looks good to me.

@gbenhaim please have a look

gbenhaim (changes requested):
IMO this issue should be solved in tektoncd/pipelines
From this comment https://github.com/tektoncd/pipeline/issues/9647#issuecomment-4126411288 it looks like there is an issue in the pipelines webhook where it doesn't load the default service account from the config.

waveywaves:
@gbenhaim We will prioritize this fix upstream as well

PR #329, opened 2026-03-31
fix: filter leaked zero-value fields from webhook admission patches

gbenhaim:
controller-runtime's CustomDefaulter re-marshals the full Go struct after Default() and diffs it against the original raw JSON. Because Go's encoding/json omitempty does not omit zero-value structs, fields like taskRunTemplate:{} and status:{} leak into the admission patch even though the webhook never modifies them. This prevents the downstream Tekton webhook from applying its own defaults (e.g. default-service-account from config-defaults).

Wrap the built-in CustomDefaulter handler with a patchFilteringWebhook that strips patches outside an explicit allowlist of fields the webhook intends to modify (/metadata/labels, /metadata/annotations, /spec/status, /spec/managedBy). The Default() function and all existing mutation logic remain unchanged.

Note: this workaround will become unnecessary if either:
- Tekton Pipeline types adopt Go 1.24's omitzero struct tag on value-type fields like TaskRunTemplate, or
- controller-runtime switches to encoding/json/v2, where omitempty correctly omits empty JSON objects ({}).

Fixes: https://github.com/konflux-ci/tekton-kueue/issues/319
Related: https://github.com/tektoncd/pipeline/issues/9647

Assisted-By: Cursor

### DIFF
--- a/internal/webhook/v1/pipelinerun_webhook.go
+++ b/internal/webhook/v1/pipelinerun_webhook.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/go-logr/logr"
 	"github.com/konflux-ci/tekton-kueue/internal/cel"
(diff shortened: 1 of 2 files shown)


pramodbindal (approved):

gbenhaim:
> > @gbenhaim I think this is a cleaner implementation yes. Could you add me as a coauthor ? I am also closing my PR.
> 
> @gbenhaim Can you please reply on this comment

done

pramodbindal:
Thanks @gbenhaim 
We also need to backport this to release 0.3.x

waveywaves:
@gbenhaim thank you for helping out with this

waveywaves:
Creating a manual cherrypick to 0.3.0

PR #377, opened 2026-04-02
Add godoc documentation to controller, webhook, and config packages

oswcab:
Add package-level and type-level documentation explaining architectural context and design decisions rather than restating what the code does:

- controller: Kueue-Tekton bridge design, synthetic resource for concurrency control, PodSets approach for quota accounting
- webhook/v1: admission flow, ConfigStore concurrency model, failurePolicy rationale
- config: field-level docs on queue assignment and multiKueue dispatch
- common: constant semantics and namespace resolution
- Fix misleading metric comments that referenced "CEL Reloads" instead of config reloads
- Remove scaffolding TODO comment

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

Contributes to: KFLUXINFRA-3247

### DIFF
--- a/internal/controller/pipelinerun_controller.go
+++ b/internal/controller/pipelinerun_controller.go
@@ -1,3 +1,14 @@
+// Package controller implements the Kueue integration for Tekton PipelineRuns.
+//
+// It bridges Tekton and Kueue by wrapping PipelineRun as a Kueue GenericJob,
+// allowing Kueue to manage PipelineRun scheduling, quota, and preemption.
+// PipelineRuns are suspended (set to Pending) by the webhook on creation and
+// only released by Kueue when cluster resources are available.
+//
+// Resource requirements for quota accounting are declared via annotations
+// on the PipelineRun (e.g. kueue.konflux-ci.dev/requests-cpu). A synthetic
+// "tekton.dev/pipelineruns" resource is always included to enable concurrency
+// limits independent of actual compute resources.
 package controller
 
 import (
@@ -36,6 +47,9 @@ import (
 // +kubebuilder:rbac:groups="tekton.dev",resources=pipelineruns/finalizers,verbs=update
 // +kubebuilder:rbac:groups="",resources=namespaces,verbs=list;watch
 
+// PipelineRun wraps tekv1.PipelineRun to implement Kueue's GenericJob and
+// JobWithCustomStop interfaces. This is a type definition (not an alias) so
+// we can attach methods without modifying the upstream Tekton type.
 type PipelineRun tekv1.PipelineRun
 
 const (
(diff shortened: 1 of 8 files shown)


internal/controller/pipelinerun_controller.go
```diff
@@ -196,6 +228,9 @@ func (p *PipelineRun) resourcesRequests() corev1.ResourceList {
 }
 
 // PodsReady implements jobframework.GenericJob.
+// This method is never called because the WaitForPodsReady configuration is
+// not enabled for PipelineRuns. Kueue tracks pod readiness for batch Jobs,
+// but PipelineRuns manage their own pod lifecycle through Tekton.
 func (p *PipelineRun) PodsReady() bool {
 	panic("pods ready shouldn't be called")
 }
```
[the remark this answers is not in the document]
oswcab: Addressed in upcoming PR, this one is only adding comments to the existing code.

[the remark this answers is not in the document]
filariow: the panic is fine here. This method is not expected to be used.

filariow (approved):

internal/webhook/v1/pipelinerun_webhook.go
```diff
@@ -71,15 +71,15 @@ func logConstructor(base logr.Logger, req *admission.Request) logr.Logger {
...
-// Kind PipelineRun when those are created or updated.
+// pipelineRunCustomDefaulter is the webhook handler that intercepts PipelineRun
+// creation requests. It suspends each PipelineRun (Pending), assigns it to a
+// Kueue queue, and applies any configured CEL mutations before the PipelineRun
+// is persisted. The webhook uses failurePolicy=fail to prevent unqueued
+// PipelineRuns from bypassing Kueue.
 //
-// NOTE: The +kubebuilder:object:generate=false marker prevents controller-gen from generating DeepCopy methods,
-// as it is used only for temporary operations and does not need to be deeply copied.
+// +kubebuilder:object:generate=false
```
filariow: Is this needed?

filariow (changes requested):

internal/webhook/v1/pipelinerun_webhook.go
oswcab: Yes, the `+kubebuilder:object:generate=false` marker was already present before this change — it prevents `controller-gen` from generating an unnecessary DeepCopy method for this struct, which isn't a Kubernetes API type. This PR just condensed the explanation that was previously a multi-line NOTE comment.

filariow: I don't see that marker [in main](https://github.com/konflux-ci/tekton-kueue/blob/fadaf9a771e3dde5cc9d0ac567f0e55feafc0605/internal/webhook/v1/pipelinerun_webhook.go#L130-L137), it seems to me it's added with this PR. Also the struct is not exported, so I wouldn't expect `controller-gen` to take it in consideration.
As I see it, it's a no-op, but I'd avoid introducing it if it's not necessary

oswcab: Removed.

filariow (approved):

PR #390, opened 2026-04-09
Fix: set response.patch to nil when all patches are filtered out

pramodbindal:
When webhook filters out all the patches then PatchType is set to nil but  there might be  Response.Patch is stil holding the value.

Setting the Response.Patch to nil will align the  response.Patch and Response.PatchType so calling validator do not fail

### DIFF
--- a/internal/webhook/v1/pipelinerun_webhook.go
+++ b/internal/webhook/v1/pipelinerun_webhook.go
@@ -84,6 +84,7 @@ func (w *patchFilteringWebhook) Handle(ctx context.Context, req admission.Reques
 	resp.Patches = resp.Patches[:n]
 	if n == 0 {
 		resp.PatchType = nil
+		resp.Patch = nil
 	}
 	return resp
 }
--- a/internal/webhook/v1/pipelinerun_webhook_test.go
+++ b/internal/webhook/v1/pipelinerun_webhook_test.go
@@ -335,6 +335,24 @@ var minimalPipelineRunJSON = []byte(`{
 	}
 }`)
 
+// prePopulatedPipelineRunJSON is a raw  admission request. It has all the fields prepopulated
+// Webhook is not supposed apply any patch here
+var prePopulatedPipelineRunJSON = []byte(`{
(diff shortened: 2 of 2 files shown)


internal/webhook/v1/pipelinerun_webhook.go
```diff
@@ -84,6 +84,7 @@ func (w *patchFilteringWebhook) Handle(ctx context.Context, req admission.Reques
 	resp.Patches = resp.Patches[:n]
 	if n == 0 {
 		resp.PatchType = nil
+		resp.Patch = nil
```
waveywaves, line 87: we should add resp.Patches = nil as well no ?

zakisk (approved):

jkhelil (approved):

filariow:
@pramodbindal may you add some details in the PR description on why we need this change? is targeting the `release-v0.3.x` intentional? any change to the unit/e2e tests to test this?

internal/webhook/v1/pipelinerun_webhook.go
gbenhaim, line 87: please add a test for this case.

gbenhaim (changes requested):

waveywaves (changes requested):

internal/webhook/v1/pipelinerun_webhook.go
pramodbindal: Its already empty. does it make any difference ?

pramodbindal: Added

waveywaves: Don't think it'll make a difference actually

waveywaves:
@pramodbindal we need to open a PR to main and then back port this maybe ? Can we do that ?

pramodbindal:
> @pramodbindal we need to open a PR to main and then back port this maybe ? Can we do that ?

https://github.com/konflux-ci/tekton-kueue/pull/392

gbenhaim (approved):

PR #391, opened 2026-04-09
Add unit tests for controller and common packages

oswcab:
Add unit tests covering all logical branches for:

- PipelineRun controller methods (GVK, IsActive, IsSuspended, Object,
  PodsReady, RestorePodSetsInfo, RunWithPodSetsInfo, Suspend, Finished,
  PodSets, resourcesRequests, Stop) with fake client and interceptors
- ConfigMapReconciler.Reconcile (not found, get error, missing key,
  invalid YAML, validation failure, success, CEL mutators)
- GetCurrentNamespace (success and file-not-found paths)
- Add an internal readNamespace(path) function that tests call directly.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

COntributes to: KFLUXINFRA-3247

### DIFF
--- a/internal/controller/configmap_controller_test.go
+++ b/internal/controller/configmap_controller_test.go
@@ -1,0 +1,168 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	v1 "github.com/konflux-ci/tekton-kueue/internal/webhook/v1"
+	"github.com/konflux-ci/tekton-kueue/pkg/common"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
+)
+
+var _ = Describe("ConfigMapReconciler", func() {
+	var (
+		reconciler *ConfigMapReconciler
+		store      *v1.ConfigStore
+		s          *runtime.Scheme
+		nsName     types.NamespacedName
+	)
+
+	BeforeEach(func() {
+		s = runtime.NewScheme()
+		Expect(clientgoscheme.AddToScheme(s)).To(Succeed())
+
+		store = &v1.ConfigStore{}
+		nsName = types.NamespacedName{
+			Name:      common.ConfigMapName,
+			Namespace: "tekton-kueue",
+		}
+	})
+
+	Describe("Reconcile", func() {
+		It("should return success when ConfigMap is not found", func(ctx context.Context) {
+			fakeClient := fake.NewClientBuilder().
+				WithScheme(s).
+				WithInterceptorFuncs(interceptor.Funcs{
+					Get: func(_ context.Context, _ client.WithWatch, key client.ObjectKey, _ client.Object, _ ...client.GetOption) error {
+						Expect(key).To(Equal(nsName))
+						return apierrors.NewNotFound(schema.GroupResource{Resource: "configmaps"}, key.Name)
+					},
+				}).
+				Build()
+			reconciler = &ConfigMapReconciler{Client: fakeClient, Store: store}
+
+			Expect(reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: nsName})).To(Equal(ctrl.Result{}))
+		})
+
+		It("should return error when Client.Get fails with a non-NotFound error", func(ctx context.Context) {
+			getErr := fmt.Errorf("connection refused")
+			fakeClient := fake.NewClientBuilder().
+				WithScheme(s).
+				WithInterceptorFuncs(interceptor.Funcs{
+					Get: func(_ context.Context, _ client.WithWatch, key client.ObjectKey, _ client.Object, _ ...client.GetOption) error {
+						Expect(key).To(Equal(nsName))
+						return getErr
+					},
+				}).
+				Build()
+			reconciler = &ConfigMapReconciler{Client: fakeClient, Store: store}
+
+			_, err := reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: nsName})
+			Expect(err).To(MatchError("connection refused"))
+		})
+
+		It("should return success when ConfigMap exists but config key is missing", func(ctx context.Context) {
+			cm := &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      common.ConfigMapName,
+					Namespace: "tekton-kueue",
+				},
+				Data: map[string]string{
+					"other-key": "some-value",
+				},
+			}
+			fakeClient := fake.NewClientBuilder().WithScheme(s).WithObjects(cm).Build()
+			reconciler = &ConfigMapReconciler{Client: fakeClient, Store: store}
+
+			Expect(reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: nsName})).To(Equal(ctrl.Result{}))
+		})
+
+		It("should requeue when config YAML is invalid", func(ctx context.Context) {
+			cm := &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      common.ConfigMapName,
+					Namespace: "tekton-kueue",
+				},
+				Data: map[string]string{
+					common.ConfigKey: "invalid: yaml: [:",
+				},
+			}
+			fakeClient := fake.NewClientBuilder().WithScheme(s).WithObjects(cm).Build()
+			reconciler = &ConfigMapReconciler{Client: fakeClient, Store: store}
+
+			result, err := reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: nsName})
+			Expect(err).To(HaveOccurred())
+			Expect(result.RequeueAfter).To(Equal(10 * time.Second))
+		})
+
+		It("should requeue when config validation fails (empty queue name)", func(ctx context.Context) {
+			cm := &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      common.ConfigMapName,
+					Namespace: "tekton-kueue",
+				},
+				Data: map[string]string{
+					common.ConfigKey: "multiKueueOverride: true",
+				},
+			}
+			fakeClient := fake.NewClientBuilder().WithScheme(s).WithObjects(cm).Build()
+			reconciler = &ConfigMapReconciler{Client: fakeClient, Store: store}
+
+			result, err := reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: nsName})
+			Expect(err).To(MatchError(ContainSubstring("queue name")))
+			Expect(result.RequeueAfter).To(Equal(10 * time.Second))
+		})
+
+		It("should update the store successfully with valid config", func(ctx context.Context) {
+			cm := &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      common.ConfigMapName,
+					Namespace: "tekton-kueue",
+				},
+				Data: map[string]string{
+					common.ConfigKey: "queueName: test-queue\nmultiKueueOverride: false",
+				},
+			}
+			fakeClient := fake.NewClientBuilder().WithScheme(s).WithObjects(cm).Build()
+			reconciler = &ConfigMapReconciler{Client: fakeClient, Store: store}
+
+			Expect(reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: nsName})).To(Equal(ctrl.Result{}))
+
+			cfg, mutators := store.GetConfigAndMutators()
+			Expect(cfg).NotTo(BeNil())
+			Expect(cfg.QueueName).To(Equal("test-queue"))
+			Expect(cfg.MultiKueueOverride).To(BeFalse())
+			Expect(mutators).To(BeEmpty())
+		})
+	})
+})
--- a/internal/controller/pipelinerun_controller.go
+++ b/internal/controller/pipelinerun_controller.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -18,6 +19,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/utils/ptr"
 	kapi "knative.dev/pkg/apis"
 
 	kueueconfig "sigs.k8s.io/kueue/apis/config/v1beta1"
@@ -143,7 +145,10 @@ func (p *PipelineRun) Object() client.Object {
 
 // PodSets implements jobframework.GenericJob.
 func (p *PipelineRun) PodSets() ([]kueue.PodSet, error) {
-	requests := p.resourcesRequests()
+	requests, err := p.resourcesRequests()
+	if err != nil {
+		return nil, err
+	}
 
 	return []kueue.PodSet{
 		{
@@ -177,22 +182,48 @@ func (p *PipelineRun) PodSets() ([]kueue.PodSet, error) {
 // By default, a resource which indicates that the workload requires 1
 // PipelineRun will be added. This is useful for controlling the number
 // of PipelineRuns that can be executed concurrently.
-//
-// WARNING: Annotations are not validated and a panic will
-// happen if they can not be parsed as `resource.Quantity`.
-func (p *PipelineRun) resourcesRequests() corev1.ResourceList {
+func (p *PipelineRun) resourcesRequests() (corev1.ResourceList, error) {
 	requests := corev1.ResourceList{
 		ResourcePipelineRunCount: resource.MustParse("1"),
 	}
 
 	for k, v := range p.GetAnnotations() {
-		if t := strings.TrimPrefix(k, annotationResourcesRequests); t != k {
-			// TODO(@filariow): how to properly validate this?
-			requests[corev1.ResourceName(t)] = resource.MustParse(v)
+		n, q, err := p.parseResourcesRequestsAnnotation(k, v)
+		switch {
+		case err != nil:
+			return nil, err
+		case n != nil && q != nil:
+			requests[*n] = *q
 		}
 	}
 
-	return requests
+	return requests, nil
+}
+
+// parseResourcesRequestsAnnotation checks if an annotation is a ResourcesRequests one.
+// It validates the extracted key and value. If the annotation is invalid the PipelineRun can not
+// be correctly processed and it needs to be fixed. To avoid a reconciliation loop an
+// UnretryableError is returned. This will tell Kueue's reconciler to avoid reconciling the
+// PipelineRun at current state again. If a new event on the PipelineRun occurs, a new
+// reconciliation will start.
+func (p *PipelineRun) parseResourcesRequestsAnnotation(k, v string) (*corev1.ResourceName, *resource.Quantity, error) {
+	t, ok := strings.CutPrefix(k, annotationResourcesRequests)
+	if !ok {
+		return nil, nil, nil
+	}
+
+	if t == "" {
+		return nil, nil, jobframework.UnretryableError(
+			fmt.Sprintf("empty resource name in annotation %s", k))
+	}
+
+	q, err := resource.ParseQuantity(v)
+	if err != nil {
+		return nil, nil, jobframework.UnretryableError(
+			fmt.Sprintf("invalid resource quantity in annotation %s=%q: %v", k, v, err))
+	}
+
+	return ptr.To(corev1.ResourceName(t)), &q, nil
 }
 
 // PodsReady implements jobframework.GenericJob.
--- a/internal/controller/pipelinerun_controller_test.go
+++ b/internal/controller/pipelinerun_controller_test.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2025.
+Copyright 2026.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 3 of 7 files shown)


pkg/common/common_utils.go
```diff
@@ -5,7 +5,7 @@ import (
 	"os"
 )
 
-const namespaceFile = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
+var namespaceFile = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
 
 func GetCurrentNamespace() (string, error) {
 	bytes, err := os.ReadFile(namespaceFile)
```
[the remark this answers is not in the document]
oswcab: Done.

internal/controller/pipelinerun_controller_test.go
```diff
@@ -17,16 +17,457 @@ limitations under the License.
...
+					"kueue.konflux-ci.dev/requests-memory": "8Gi",
+				}
+			})
+			podSets, err := p.PodSets()
+			Expect(err).NotTo(HaveOccurred())
+
+			requests := podSets[0].Template.Spec.Containers[0].Resources.Requests
+			Expect(requests).To(HaveKeyWithValue(corev1.ResourceName("cpu"), resource.MustParse("4")))
+			Expect(requests).To(HaveKeyWithValue(corev1.ResourceName("memory"), resource.MustParse("8Gi")))
+			Expect(requests).To(HaveKeyWithValue(
```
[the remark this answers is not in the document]
oswcab: Done.

oswcab:
> `GetCurrentNamespace` now trims whitespace and returns a wrapped error including the file path. Verify no downstream code relied on the previous untrimmed namespace (including newline) behavior, and that callers comparing errors are not broken by the new wrapped error string.



> ## PR Reviewer Guide 🔍
> #### (Review updated until commit [0a1bfc3](https://github.com/konflux-ci/tekton-kueue/commit/0a1bfc37498465d192d4e6d06b70d0e429dfb387))
> Here are some key observations to aid the review process:
> 
> ⏱️ **Estimated effort to review**: 3 🔵🔵🔵⚪⚪
> 🧪 **PR contains tests**
> 🔒 **No security concerns identified**
> ⚡ **Recommended focus areas for review**
> 
> [**API Change**](https://github.com/konflux-ci/tekton-kueue/pull/391/files#diff-cf207ef316b1e5bd5fa337aef509287e0d083d5aca8a2968e90aaf92df13518dR145-R200)
> `resourcesRequests` now returns `(corev1.ResourceList, error)` and `PodSets` propagates that error. Ensure all other call sites (outside this diff) are updated to handle the new error return and that any interface expectations for `PodSets()` returning errors are still satisfied across the controller/jobframework integration.
> 
> ```go
> // PodSets implements jobframework.GenericJob.
> func (p *PipelineRun) PodSets() ([]kueue.PodSet, error) {
> 	requests, err := p.resourcesRequests()
> 	if err != nil {
> 		return nil, err
> 	}
> 
> 	return []kueue.PodSet{
> 		{
> 			Name: "pod-set-1",
> 			Template: corev1.PodTemplateSpec{
> 				Spec: corev1.PodSpec{
> 					Containers: []corev1.Container{
> 						{
> 							Name:  "dummy",
> 							Image: "dummy",
> 							Resources: corev1.ResourceRequirements{
> 								Requests: requests,
> 							},
> 						},
> 					},
> 				},
> 			},
> 			Count: 1,
> 		},
> 	}, nil
> }
> 
> // resourcesRequests will match all annotations starting with
> // `kueue.konflux-ci.dev/requests-`. Valid annotations to set
> // the requested resources are then:
> // * `kueue.konflux-ci.dev/requests-cpu`
> // * `kueue.konflux-ci.dev/requests-memory`
> // * `kueue.konflux-ci.dev/requests-storage`
> // * `kueue.konflux-ci.dev/requests-ephemeral-storage`
> //
> // By default, a resource which indicates that the workload requires 1
> // PipelineRun will be added. This is useful for controlling the number
> // of PipelineRuns that can be executed concurrently.
> func (p *PipelineRun) resourcesRequests() (corev1.ResourceList, error) {
> 	requests := corev1.ResourceList{
> 		ResourcePipelineRunCount: resource.MustParse("1"),
> 	}
> 
> 	for k, v := range p.GetAnnotations() {
> 		if t, ok := strings.CutPrefix(k, annotationResourcesRequests); ok {
> 			q, err := resource.ParseQuantity(v)
> 			if err != nil {
> 				return nil, fmt.Errorf("invalid resource quantity in annotation %s=%q: %w", k, v, err)
> 			}
> 			requests[corev1.ResourceName(t)] = q
> 		}
> 	}
> 
> 	return requests, nil
> }
> ```
> 
> [**Behavior Change**](https://github.com/konflux-ci/tekton-kueue/pull/391/files#diff-cf207ef316b1e5bd5fa337aef509287e0d083d5aca8a2968e90aaf92df13518dR184-R199)
> Invalid resource quantity annotations now cause reconciliation flow to error rather than panic. Confirm this is the desired operational behavior (e.g., how the controller surfaces this and whether it should be a hard error vs. ignoring invalid annotations), and that error messages are stable/appropriate for logs and users.
> 
> ```go
> func (p *PipelineRun) resourcesRequests() (corev1.ResourceList, error) {
> 	requests := corev1.ResourceList{
> 		ResourcePipelineRunCount: resource.MustParse("1"),
> 	}
> 
> 	for k, v := range p.GetAnnotations() {
> 		if t, ok := strings.CutPrefix(k, annotationResourcesRequests); ok {
> 			q, err := resource.ParseQuantity(v)
> 			if err != nil {
> 				return nil, fmt.Errorf("invalid resource quantity in annotation %s=%q: %w", k, v, err)
> 			}
> 			requests[corev1.ResourceName(t)] = q
> 		}
> 	}
> 
> 	return requests, nil
> ```
> 
> [**Error Handling**](https://github.com/konflux-ci/tekton-kueue/pull/391/files#diff-0e99f6969410c40bea8614c28f987101f06cdf68f9d7bfaf70d7d50a6c01575dR11-R20)
> `GetCurrentNamespace` now trims whitespace and returns a wrapped error including the file path. Verify no downstream code relied on the previous untrimmed namespace (including newline) behavior, and that callers comparing errors are not broken by the new wrapped error string.
> 
> ```go
> func GetCurrentNamespace() (string, error) {
> 	return readNamespace(namespaceFile)
> }
> 
> func readNamespace(path string) (string, error) {
> 	bytes, err := os.ReadFile(path)
> 	if err != nil {
> 		return "", fmt.Errorf("not able to read namespace file %s: %w", path, err)
> 	}
> 	return strings.TrimSpace(string(bytes)), nil
> ```

---

1. resourcesRequests / PodSets: resourcesRequests is private, only called from PodSets(). The kueue GenericJob interface already defines PodSets() ([]kueue.PodSet, error) and the framework propagates errors through reconciliation with requeue —  no call sites need updating.
2. Invalid annotation behavior: The error from PodSets() propagates to controller-runtime's reconciler, which logs it and
  requeues with exponential backoff. This is the right behavior — the PipelineRun stays pending until the annotation is fixed, and the error message (invalid resource quantity in annotation kueue.../requests-cpu="not-a-quantity") is clear enough for operators to diagnose. A panic would crash the entire controller, affecting all PipelineRuns.

 3. GetCurrentNamespace: Both callers (cmd/main.go:269, configmap_controller.go:23) use if err != nil — they check the error, never compare the string. The namespace is used for cache scoping (line 282) and predicate filtering (line 31), both exact string comparisons that would have broken with a trailing newline. So TrimSpace is a fix, not a behavior change that could break anything.

filariow:
this PR is introducing a considerate amount of changes with different scopes. To simplify the review process, may we split it in multiple smaller and more scoped PRs?

>     * PipelineRun controller methods (GVK, IsActive, IsSuspended, Object,
>       PodsReady, RestorePodSetsInfo, RunWithPodSetsInfo, Suspend, Finished,
>       PodSets, resourcesRequests, Stop) with fake client and interceptors
>     * ConfigMapReconciler.Reconcile (not found, get error, missing key,
>       invalid YAML, validation failure, success, CEL mutators)
>     * GetCurrentNamespace (success and file-not-found paths)
>     * Add an internal readNamespace(path) function that tests call directly.

oswcab:
> this PR is introducing a considerate amount of changes with different scopes. To simplify the review process, may we split it in multiple smaller and more scoped PRs?
> 
> > ```
> > * PipelineRun controller methods (GVK, IsActive, IsSuspended, Object,
> >   PodsReady, RestorePodSetsInfo, RunWithPodSetsInfo, Suspend, Finished,
> >   PodSets, resourcesRequests, Stop) with fake client and interceptors
> > * ConfigMapReconciler.Reconcile (not found, get error, missing key,
> >   invalid YAML, validation failure, success, CEL mutators)
> > * GetCurrentNamespace (success and file-not-found paths)
> > * Add an internal readNamespace(path) function that tests call directly.
> > ```

This is the list of tests that were added/ Do you want one PR per test file? Otherwise, what I see we can extract to a different PR is the change done in `pkg/common/common_utils.go` that was done just to facilitate the testing.

internal/controller/configmap_controller_test.go
```diff
@@ -0,0 +1,228 @@
...
+	var (
+		testCtx    context.Context
+		reconciler *ConfigMapReconciler
+		store      *v1.ConfigStore
+		s          *runtime.Scheme
+		nsName     types.NamespacedName
+	)
+
+	BeforeEach(func() {
+		testCtx = context.Background()
```
sadlerap: I don't think you need to initialize your own context.  You can have ginkgo inject one as a parameter to the closures passed to `It` clauses.

internal/controller/configmap_controller_test.go
```diff
@@ -0,0 +1,228 @@
...
+				Data: map[string]string{
+					"other-key": "some-value",
+				},
+			}
+			fakeClient := fake.NewClientBuilder().WithScheme(s).WithObjects(cm).Build()
+			reconciler = &ConfigMapReconciler{Client: fakeClient, Store: store}
+
+			result, err := reconciler.Reconcile(testCtx, ctrl.Request{NamespacedName: nsName})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(result).To(Equal(ctrl.Result{}))
```
sadlerap: In general, you can simplify some of your checks to look like this:

```suggestion
			Expect(reconciler.Reconcile(testCtx, ctrl.Request{NamespacedName: nsName})).To(Equal(ctrl.Result{}))
```

@oswcab Up to you if you if you want to apply this or not.

internal/controller/pipelinerun_controller_test.go
```diff
@@ -17,16 +17,480 @@ limitations under the License.
...
+		It("should return error when annotation has invalid resource quantity", func() {
+			p := newTestPipelineRun(func(plr *tekv1.PipelineRun) {
+				plr.Annotations = map[string]string{
+					"kueue.konflux-ci.dev/requests-cpu": "not-a-quantity",
+				}
+			})
+			podSets, err := p.PodSets()
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("invalid resource quantity"))
+			Expect(podSets).To(BeNil())
```
sadlerap: Ditto for error cases, though you'll need to add a `.Error()` invocation:

```suggestion
			Expect(p.PodSets()).Error().To(ContainSubstring("invalid resource quantity"))
```

internal/controller/configmap_controller_test.go
```diff
@@ -0,0 +1,228 @@
...
+
+			cfg, mutators := store.GetConfigAndMutators()
+			Expect(cfg.QueueName).To(Equal("test-queue"))
+			Expect(mutators).To(HaveLen(1))
+			Expect(mutators[0]).To(BeAssignableToTypeOf(&cel.CELMutator{}))
+		})
+	})
```
filariow: IIUC, this [logic is enforced by the store](https://github.com/konflux-ci/tekton-kueue/blob/1f9856dd689e7e1625bcbe87d63d25d9515eac81/internal/webhook/v1/config_store.go#L37-L56), not by the Reconciler. Should we move it under the [Store's tests](https://github.com/konflux-ci/tekton-kueue/blob/main/internal/webhook/v1/config_store_test.go)?

internal/controller/pipelinerun_controller.go
```diff
@@ -177,22 +181,22 @@ func (p *PipelineRun) PodSets() ([]kueue.PodSet, error) {
...
 	requests := corev1.ResourceList{
 		ResourcePipelineRunCount: resource.MustParse("1"),
 	}
 
 	for k, v := range p.GetAnnotations() {
-		if t := strings.TrimPrefix(k, annotationResourcesRequests); t != k {
-			// TODO(@filariow): how to properly validate this?
-			requests[corev1.ResourceName(t)] = resource.MustParse(v)
+		if t, ok := strings.CutPrefix(k, annotationResourcesRequests); ok {
+			q, err := resource.ParseQuantity(v)
```
filariow: I don't recall why I decided to rely on `MustParse` when I implemented this :/

pkg/common/common_utils_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+package common
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestReadNamespace(t *testing.T) {
```
filariow: I think it would be preferable to use ginkgo here too

filariow (commented):
still midway, some feedback! thanks for working on this :)

internal/controller/configmap_controller_test.go
oswcab: Good point, fixed.

internal/controller/configmap_controller_test.go
oswcab: Done.

internal/controller/configmap_controller_test.go
oswcab: Makes sense, moved.

internal/controller/pipelinerun_controller_test.go
oswcab: Done.

pkg/common/common_utils_test.go
oswcab: Done.

internal/controller/pipelinerun_controller.go
oswcab: Added a small comment with the rationale behind the change.

filariow: I think I recalled. This annotation is set by `tekton-kueue` itself. This is not expected to happen in the real world. If that happens it should be catch in tests. A panic seemed a good way to break instantaneously, whereas an error could still be managed/suppressed.

internal/controller/pipelinerun_controller_test.go
```diff
@@ -17,16 +17,477 @@ limitations under the License.
...
+	"sigs.k8s.io/kueue/pkg/controller/jobframework"
+	"sigs.k8s.io/kueue/pkg/podset"
 )
 
-var _ = Describe("PipelineRun Controller", func() {
-	Context("When reconciling a resource", func() {
+func newTestPipelineRun(opts ...func(*tekv1.PipelineRun)) *PipelineRun {
+	plr := &tekv1.PipelineRun{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "tekton.dev/v1",
```
filariow: minor: if you prefer we can use the info in the referenced tekton package


```suggestion
			APIVersion: tekv1.SchemeGroupVersion.String(),
```

internal/controller/pipelinerun_controller_test.go
```diff
@@ -17,16 +17,477 @@ limitations under the License.
...
+}
+
+var _ = Describe("PipelineRun", func() {
+
+	Describe("GVK", func() {
+		It("should return the PipelineRun GroupVersionKind", func() {
+			p := newTestPipelineRun()
+			gvk := p.GVK()
+			Expect(gvk.Group).To(Equal("tekton.dev"))
+			Expect(gvk.Version).To(Equal("v1"))
```
filariow: ditto

internal/controller/pipelinerun_controller_test.go
```diff
@@ -17,16 +17,477 @@ limitations under the License.
...
+				p := newTestPipelineRun(func(plr *tekv1.PipelineRun) {
+					plr.Spec.Status = st
+				})
+				err := p.RunWithPodSetsInfo(nil)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(p.Spec.Status).To(BeEmpty(), "for initial status %q", st)
+			}
+		})
+	})
+
```
filariow: I think a DescribeTable is a better fit here

```suggestion
	DescribeTable("RunWithPodSetsInfo should clear Spec.Status regardless of prior value", func(initialStatus string) {
		p := newTestPipelineRun(func(plr *tekv1.PipelineRun) {
			plr.Spec.Status = tekv1.PipelineRunSpecStatus(initialStatus)
		})
		err := p.RunWithPodSetsInfo(nil)
		Expect(err).NotTo(HaveOccurred())
		Expect(p.Spec.Status).To(BeEmpty(), "for initial status %q", initialStatus)
	},
		Entry("Empty", ""),
		Entry(tekv1.PipelineRunSpecStatusPending, tekv1.PipelineRunSpecStatusPending),
		Entry(tekv1.PipelineRunSpecStatusStoppedRunFinally, tekv1.PipelineRunSpecStatusStoppedRunFinally),
	)

```

internal/controller/pipelinerun_controller_test.go
```diff
@@ -17,16 +17,477 @@ limitations under the License.
...
+	Describe("resourcesRequests", func() {
+		It("should return only the default pipelinerun count when no annotations are set", func() {
+			p := newTestPipelineRun()
+			requests, err := p.resourcesRequests()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(requests).To(HaveLen(1))
+			Expect(requests).To(HaveKeyWithValue(
+				corev1.ResourceName(ResourcePipelineRunCount),
+				resource.MustParse("1"),
+			))
```
filariow: minor (cosmetic): if you like it more you can also write this with an `And`

```suggestion
			Expect(requests).To(
				And(
					HaveLen(1),
					HaveKeyWithValue(
						corev1.ResourceName(ResourcePipelineRunCount),
						resource.MustParse("1"),
					),
			))
```

internal/controller/pipelinerun_controller_test.go
```diff
@@ -17,16 +17,477 @@ limitations under the License.
...
+					"kueue.konflux-ci.dev/requests-cpu": "not-a-quantity",
+				}
+			})
+			Expect(p.PodSets()).Error().To(MatchError(ContainSubstring("invalid resource quantity")))
+		})
+	})
+
+	Describe("Stop", func() {
+		var (
+			testCtx context.Context
```
filariow: we should remove this

internal/controller/pipelinerun_controller_test.go
```diff
@@ -17,16 +17,477 @@ limitations under the License.
...
+			})
+			requests, err := p.resourcesRequests()
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("invalid resource quantity"))
+			Expect(err.Error()).To(ContainSubstring("not-a-quantity"))
+			Expect(requests).To(BeNil())
+		})
+	})
+
+	Describe("PodSets", func() {
```
filariow, line 346: we should also check that it panics/returns an error (it depends from the other conversation) when the annotation key is malformed

filariow (changes requested):

internal/controller/pipelinerun_controller_test.go
oswcab: Done.

internal/controller/pipelinerun_controller_test.go
oswcab: Done.

internal/controller/pipelinerun_controller_test.go
oswcab: Done.

internal/controller/pipelinerun_controller_test.go
oswcab: Done.

internal/controller/pipelinerun_controller_test.go
oswcab: Done.

internal/controller/pipelinerun_controller_test.go
oswcab: Done.

internal/controller/pipelinerun_controller.go
oswcab: I still think using an error is a better option. I think treating annotations as trusted can be fragile, as it is set now by `tekton-kueue` but, given annotations are mutable, that can change if ever an operator or a webhook modify them in a future change. Besides, a panic, if ever it happens, would crash the complete process, affecting reconciliation, while an error would reject the bad pipelinerun while the controller keeps managing all others. The other advantage could be observability, IMO, as the error would surface in the controller log and in the pipelinerun status also. And last, k8s controllers are expected to handle bad input gracefully, not to panic. 

And finally, you have a good point: If this should never happen then the error path will never be executed in production. But if it ever happens, a panic will crash the controller, which is not a good option.

filariow: in controller-runtime controllers, panics are recovered by default. When a panic happens, we'll see an error logged in the controller logs. I would expect this to happen here too. A panic could be more secure as reconciliation logic will not handle it, but controller's logic will.

filariow: agreed to propagate an [UnretryableError](https://github.com/kubernetes-sigs/kueue/blob/2f2a452481b23fd664e4133ad51de1caa1d49c3a/pkg/controller/jobframework/errors.go#L23) to avoid reconciliation loops. Indeed, the Kueue's Reconciler will log the UnretryableError but [won't propagate it](https://github.com/kubernetes-sigs/kueue/blob/2f2a452481b23fd664e4133ad51de1caa1d49c3a/pkg/controller/jobframework/reconciler.go#L259). This will avoid reconciliation loops for malformed values.

It's not appropriate to test this in unit tests, we'll introduce e2e tests to check this scenario

internal/controller/pipelinerun_controller.go
```diff
@@ -177,22 +181,31 @@ func (p *PipelineRun) PodSets() ([]kueue.PodSet, error) {
...
+					fmt.Sprintf("empty resource name in annotation %s", k))
+			}
+			// ParseQuantity instead of MustParse: annotation values are user-supplied
+			// input that could be malformed, so we return an error instead of panicking.
+			q, err := resource.ParseQuantity(v)
+			if err != nil {
+				return nil, jobframework.UnretryableError(
+					fmt.Sprintf("invalid resource quantity in annotation %s=%q: %v", k, v, err))
+			}
+			requests[corev1.ResourceName(t)] = q
```
filariow: It feels like it's worth to move this to an auxiliary function. Something like the following:

```go
import 	"k8s.io/utils/ptr"

// parseResourcesRequestsAnnotation checks if an annotation is a ResourcesRequests one.
// It validates the extracted key and value. If the annotation is invalid the PipelineRun can not
// correctly processed an it needs to be fixed. To avoid a reconciliation loop an
// UnretryableError is returned. This will tell the Kueue's reconciler to avoid reconciling the
// PipelineRun at current state again. If a new event on the PipelineRun occurs, a new
// Reconciliation will start.
func (p *PipelineRun) parseResourcesRequestsAnnotation(k, v string) (*corev1.ResourceName, *resource.Quantity, error) {
	t, ok := strings.CutPrefix(k, annotationResourcesRequests)
	if !ok {
		// not a ResourcesRequest Annotation
		return nil, nil, nil
	}

	// validate annotation's key
	if t == "" {
		// Annotation validation errors won't resolve on retry — mark as
		// unretryable to prevent the reconciler from looping.
		return nil, nil, jobframework.UnretryableError(
			fmt.Sprintf("empty resource name in annotation %s", k))
	}

	// parse and validate quantity in annotation's value
	q, err := resource.ParseQuantity(v)
	if err != nil {
		return nil, nil, jobframework.UnretryableError(
			fmt.Sprintf("invalid resource quantity in annotation %s=%q: %v", k, v, err))
	}

	return ptr.To(corev1.ResourceName(t)), &q, nil
}
```

```suggestion
		n, q, err := p.parseResourcesRequestsAnnotation(k, v)
		switch {
		case err != nil:
			return nil, err
		case n != nil && q != nil:
			requests[*n] = *q
```

What do you think?

filariow (changes requested):

internal/controller/pipelinerun_controller.go
oswcab: Good, thanks.

filariow (approved):
thanks for working on this! 🎉

PR #400, opened 2026-04-17
accept QPS and Burst from flags

filariow:
We want to be able to configure controller-runtime Client's QPS and Burst by changing the deployment manifest only.

This change will make tekton-kueue get these values from the command flags. In their absence, controller-runtime default values will be used.

Signed-off-by: Francesco Ilario <filario@redhat.com>

### DIFF
--- a/cmd/cmd_suite_test.go
+++ b/cmd/cmd_suite_test.go
@@ -1,4 +1,4 @@
-package main_test
+package main
 
 import (
 	"testing"
--- a/cmd/cmd_suite_test.go
+++ b/cmd/cmd_suite_test.go
@@ -1,0 +1,13 @@
+package main_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestCmd(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Cmd Suite")
+}
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -20,6 +20,7 @@ import (
 	"crypto/tls"
 	"flag"
 	"fmt"
+	"math"
 	"os"
 	"path/filepath"
 	"reflect"
@@ -77,6 +78,8 @@ type SharedFlags struct {
 	ProbeAddr       string
 	EnableHTTP2     bool
 	ZapOptions      *zap.Options
+	QPS             float64
+	Burst           int
 }
 
 func (s *SharedFlags) AddFlags(fs *flag.FlagSet) {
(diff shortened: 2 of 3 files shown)


cmd/main.go
```diff
@@ -93,6 +96,8 @@ func (s *SharedFlags) AddFlags(fs *flag.FlagSet) {
 	fs.StringVar(&s.ProbeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	fs.BoolVar(&s.EnableHTTP2, "enable-http2", false,
 		"If set, HTTP/2 will be enabled for the metrics and webhook servers")
+	fs.Float64Var(&s.QPS, "qps", 0, "The Kubernetes API Client's QPS value")
+	fs.IntVar(&s.Burst, "burst", 0, "The Kubernetes API Client's Burst value")
```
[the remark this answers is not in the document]
filariow: good one, fixed

cmd/main_ginkgo_test.go
```diff
@@ -0,0 +1,37 @@
...
+	DescribeTable("returns default value if provided value is not a valid float32",
+		func(v float64) {
+			a, ok := safeFloat64ToFloat32OrZero(v)
+			Expect(ok).To(BeFalse(), "provided value '%v' is recognized as legit", v)
+			Expect(a).To(Equal(float32(0)))
+		},
+		Entry("Above Max Float32", math.Nextafter(float64(math.MaxFloat32), math.MaxFloat64)),
+		Entry("Below Min Float32", -(math.Nextafter(float64(math.MaxFloat32), math.MaxFloat64))),
+		Entry("Max Float64", float64(math.MaxFloat64)),
+		Entry("Min Float64", float64(-math.MaxFloat64)),
```
sadlerap, line 35: could you add entries for `inf` and `-inf` as well as `NaN`?  AIUI those values can pass our CLI parser, and we will need to filter them out.

filariow: done

cmd/cmd_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package main_test
```
filariow:
```suggestion
package main
```

cmd/main_ginkgo_test.go
```diff
@@ -0,0 +1,40 @@
+package main
+
+import (
+	"math"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("safeFloat64ToFloat32OrDefault", func() {
```
filariow:
```suggestion
var _ = Describe("safeFloat64ToFloat32OrZero", func() {
```

cmd/main.go
```diff
@@ -319,6 +350,13 @@ func runWebhook(args []string) {
 	}
 }
 
+func safeFloat64ToFloat32OrZero(v float64) (float32, bool) {
+	if math.IsNaN(v) || math.IsInf(v, 0) || v < -math.MaxFloat32 || v > math.MaxFloat32 {
+		return 0, false
+	}
+	return float32(v), true
```
sadlerap, line 357: I think it might be semantically clearer if we first attempt to narrow a f64 to a f32 then check for invalid values:


```suggestion
	s := float32(v)
	if math.IsNaN(s) || math.IsInf(s, 0) {
		return 0, false
	}
	return s, true
```

WDYT?

sadlerap (commented):
@filariow one suggestion for your consideration, lgtm otherwise

cmd/main.go
filariow: `IsNan` and `IsInf` wants `float64`. There's no util for `float32`. I also tried the following version with no success.

```
func safeFloat64ToFloat32OrZero(v float64) (float32, bool) {
	s := float64(float32(v))
	if math.IsNaN(s) || math.IsInf(s, 0) {
		return 0, false
	}
	return float32(v), true
}
```

sadlerap (approved):

PR #416, opened 2026-04-30
KONFLUX-13356: fix CVE-2026-40938 by bumping tektoncd/pipeline to v1.11.1

glevi-rh:
## What
Fix [CVE-2026-40938](https://nvd.nist.gov/vuln/detail/CVE-2026-40938): The git resolver's revision parameter is passed directly to `git fetch` without validation, allowing injection of arbitrary flags like `--upload-pack=<binary>`, leading to arbitrary code execution and cluster-wide secret exfiltration.

[KONFLUX-13356](https://redhat.atlassian.net/browse/KONFLUX-13356)

- Bump `tektoncd/pipeline` v1.10.2 → v1.11.1
- Bump `sigs.k8s.io/kueue` v0.14.8 → v0.16.6 (required for k8s.io v0.35+ compat)
- Adapt code to kueue v0.16 API (v1beta1 → v1beta2)
- Update Renovate version ceilings

### Notable code changes

- **GenericJob interface methods** (`Finished`, `PodSets`, `PodsReady`, `RunWithPodSetsInfo`): added `context.Context` parameter, required by kueue v0.16's `jobframework.GenericJob` interface.
- **WaitForPodsReady**: in kueue config v1beta2, the `Enable` bool field was removed — a non-nil `*WaitForPodsReady` pointer now means the feature is enabled. Changed to pass `nil` to keep it disabled, matching the original v1beta1 behavior where `Enable` defaulted to `false`.
- **PriorityClassName → PriorityClassRef**: kueue v1beta2 replaced the `PriorityClassName` string field with a `PriorityClassRef` struct (`Group`, `Kind`, `Name`). Updated e2e test assertions accordingly.
- **v1beta1 → v1beta2**: updated all Go imports (`kueue/apis/kueue/v1beta2`, scheme registration), YAML manifests (`apiVersion: kueue.x-k8s.io/v1beta2`), and e2e tests. `config.kueue.x-k8s.io/v1beta1` (kueue operator Configuration API) left unchanged — still valid in v0.16.

## Why
The CVE fix is only in tektoncd/pipeline v1.11.1+, which requires k8s.io v0.35+, which requires kueue v0.16+.

## Validation
- Build passes locally
- [x]  E2e validation on a live cluster needed before merge

## Risk Assessment
**Risk Level:** Medium
**Rollback:** Revert PR and redeploy previous image
Kueue API moves from v1beta1 to v1beta2. The deployed kueue operator already supports v1beta2 natively, but e2e validation is recommended.

[KONFLUX-13356]: https://redhat.atlassian.net/browse/KONFLUX-13356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


go.mod
```diff
@@ -9,16 +9,16 @@ require (
...
+	k8s.io/apimachinery v0.35.3
+	k8s.io/client-go v0.35.3
 	k8s.io/klog/v2 v2.140.0
 	k8s.io/utils v0.0.0-20260319190234-28399d86e0b5
-	knative.dev/pkg v0.0.0-20260120122510-4a022ed9999a
-	sigs.k8s.io/controller-runtime v0.22.3
-	sigs.k8s.io/kueue v0.14.8
+	knative.dev/pkg v0.0.0-20260318013857-98d5a706d4fd
+	sigs.k8s.io/controller-runtime v0.22.4
+	sigs.k8s.io/kueue v0.16.6
```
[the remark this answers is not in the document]
glevi-rh: The issue is invalid. `hack/kueue-config.yaml` is a ConfigMap with an embedded YAML string — `kubectl apply` doesn't validate the inner apiVersion. Kueue v0.16.6 still supports `config.kueue.x-k8s.io/v1beta1` internally (deprecated, removed in v0.18). No current breakage, no action needed for this PR.

internal/controller/pipelinerun_controller_test.go
sadlerap, line 1: Rather than obtaining a context from `context.TODO()`, you can instead get a context from ginkgo by adding a `context.Context` parameter to the closures passed to `It` clauses.  We do it elsewhere in this test suite, and I think it would be good to be consistent.

internal/controller/pipelinerun_controller.go
```diff
@@ -95,7 +95,7 @@ func SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
 		mgr.GetClient(),
 		mgr.GetFieldIndexer(),
 		mgr.GetEventRecorderFor("kueue-plr"),
-		jobframework.WithWaitForPodsReady(&kueueconfig.WaitForPodsReady{}),
+		jobframework.WithWaitForPodsReady((*kueueconfig.WaitForPodsReady)(nil)),
```
sadlerap, line 101: It would be good to have a comment explaining why this needs to be a nil pointer.

```suggestion
		// In v1beta2, WaitForPodsReady removed its Enable bool field — a non-nil
		// pointer now means the feature is enabled. Pass nil to keep it disabled,
		// matching the original v1beta1 behavior where Enable defaulted to false.
		jobframework.WithWaitForPodsReady((*kueueconfig.WaitForPodsReady)(nil)),
```

sadlerap (approved):

gbenhaim (changes requested):
We need to check first if we can upgrade the kueue version Konflux is using.

glevi-rh:
> We need to check first if we can upgrade the kueue version Konflux is using.

How can we check it?

gbenhaim (approved):

PR #511, opened 2026-07-24
fix: set initial Pending status condition on queued PipelineRuns

ankrsinha:
## What

Sets an initial Succeeded=Unknown / Reason=`PipelineRunPending` status condition on PipelineRuns that are queued by tekton-kueue.

A lightweight status reconciler is registered alongside the existing Kueue workload reconciler in `SetupWithManager`. When it sees a PipelineRun with` spec.status=PipelineRunPending` and empty `status.conditions`, it writes the pending condition via the status subresource.

## Why

The tekton-kueue webhook sets `spec.managedBy=kueue.x-k8s.io/multikueue` on every PipelineRun. Tekton's own controller filters out any PipelineRun whose `managedBy` is not `tekton.dev/pipeline`, so it never reconciles these runs and never populates status.conditions.

Since `kubectl get pipelinerun`, `tkn`/`opc`, and the OpenShift Console UI all read `status.conditions` to display status, queued PipelineRuns show a blank/--- status instead of Pending - making it look like the run is broken rather than waiting for admission.

## Changes

- **internal/controller/pipelinerun_controller.go**: Added `pipelineRunStatusReconciler` and registered it in `SetupWithManager`. Added RBAC marker for pipelineruns/status.
- **internal/controller/pipelinerun_controller_test.go**: Tests covering all reconciler paths (not found, not pending, already has condition, sets condition, update failure).
- **config/rbac/role.yaml**: Auto-generated - adds patch/update on pipelineruns/status.

## Verification

- [x] `make lint` passes
- [x] `make test` passes
- [x] Tested with `tekton-kueue mutate` CLI (if CEL/webhook changes)
- [x] `make manifests` and `make generate` run cleanly (if API/RBAC changes)

### DIFF
--- a/internal/controller/pipelinerun_status_reconciler.go
+++ b/internal/controller/pipelinerun_status_reconciler.go
@@ -1,0 +1,81 @@
+package controller
+
+import (
+	"context"
+
+	"github.com/konflux-ci/tekton-kueue/pkg/common"
+	tekv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kapi "knative.dev/pkg/apis"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+// +kubebuilder:rbac:groups="tekton.dev",resources=pipelineruns/status,verbs=update;patch
+
+// pipelineRunStatusReconciler watches PipelineRuns and sets a Succeeded=Unknown
+// condition with reason "PipelineRunPending" on runs that are suspended
+// (spec.status=PipelineRunPending) but have no status.conditions yet. Tekton's
+// controller skips PipelineRuns with spec.managedBy set, so without this
+// reconciler the status stays blank in CLI and UI tools until MultiKueue admits
+// the workload and dispatches it to a spoke cluster.
+type pipelineRunStatusReconciler struct {
+	client client.Client
+}
+
+func (r *pipelineRunStatusReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := ctrl.LoggerFrom(ctx)
+
+	var plr tekv1.PipelineRun
+	if err := r.client.Get(ctx, req.NamespacedName, &plr); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	if plr.Spec.ManagedBy == nil || *plr.Spec.ManagedBy != common.ManagedByMultiKueueLabel {
+		return ctrl.Result{}, nil
+	}
+
+	if plr.Spec.Status != tekv1.PipelineRunSpecStatusPending {
+		return ctrl.Result{}, nil
+	}
+
+	if plr.Status.GetCondition(kapi.ConditionSucceeded) != nil {
+		return ctrl.Result{}, nil
+	}
+
+	plr.Status.SetCondition(&kapi.Condition{
+		Type:               kapi.ConditionSucceeded,
+		Status:             corev1.ConditionUnknown,
+		Reason:             "PipelineRunPending",
+		Message:            "PipelineRun is pending, waiting for MultiKueue admission",
+		LastTransitionTime: kapi.VolatileTime{Inner: metav1.Now()},
+	})
+
+	log.V(2).Info("Setting PipelineRunPending status condition")
+	return ctrl.Result{}, r.client.Status().Update(ctx, &plr)
+}
+
+// needsPendingStatusCondition returns true for PipelineRuns that are
+// MultiKueue-managed, pending, and likely missing a status condition.
+func needsPendingStatusCondition(o client.Object) bool {
+	plr, ok := o.(*tekv1.PipelineRun)
+	if !ok {
+		return false
+	}
+	return plr.Spec.ManagedBy != nil &&
+		*plr.Spec.ManagedBy == common.ManagedByMultiKueueLabel &&
+		plr.Spec.Status == tekv1.PipelineRunSpecStatusPending
+}
+
+// SetupPendingStatusWithManager registers the pending-status reconciler with the manager.
+func SetupPendingStatusWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		Named("PipelineRunStatus").
+		For(&tekv1.PipelineRun{}, builder.WithPredicates(
+			predicate.NewPredicateFuncs(needsPendingStatusCondition),
+		)).
+		Complete(&pipelineRunStatusReconciler{client: mgr.GetClient()})
+}
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -256,6 +256,11 @@ func runController(args []string) {
 		os.Exit(1)
 	}
 
+	if err = controller.SetupPendingStatusWithManager(mgr); err != nil {
+		setupLog.Error(err, "Failed to setup pending status controller")
+		os.Exit(1)
+	}
+
 	err = controller.SetupIndexer(ctx, mgr.GetFieldIndexer())
 	if err != nil {
 		setupLog.Error(err, "Failed to setup the indexer")
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -78,3 +78,10 @@ rules:
   - pipelineruns/finalizers
   verbs:
   - update
+- apiGroups:
+  - tekton.dev
+  resources:
+  - pipelineruns/status
+  verbs:
+  - patch
+  - update
--- a/internal/controller/pipelinerun_status_reconciler_test.go
+++ b/internal/controller/pipelinerun_status_reconciler_test.go
@@ -1,0 +1,285 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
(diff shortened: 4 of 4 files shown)


internal/controller/pipelinerun_controller.go
```diff
@@ -284,3 +296,42 @@ func (p *PipelineRun) RunWithPodSetsInfo(_ context.Context, podSetsInfo []podset
...
 	// Not implemented because this is not called when JobWithCustomStop is implemented.
 }
+
+// pipelineRunStatusReconciler watches PipelineRuns and sets a Succeeded=Unknown
+// condition with reason "PipelineRunPending" on runs that are suspended
+// (spec.status=PipelineRunPending) but have no status.conditions yet. Tekton's
+// controller skips PipelineRuns with spec.managedBy set, so without this
+// reconciler the status stays blank in CLI and UI tools until Kueue admits
+// the workload and dispatches it to a spoke cluster.
+type pipelineRunStatusReconciler struct {
```
pramodbindal: Please keep this reconciler in its own file.

internal/controller/pipelinerun_controller.go
```diff
@@ -284,3 +296,42 @@ func (p *PipelineRun) RunWithPodSetsInfo(_ context.Context, podSetsInfo []podset
...
+
+	if plr.Status.GetCondition(kapi.ConditionSucceeded) != nil {
+		return ctrl.Result{}, nil
+	}
+
+	plr.Status.Conditions = append(plr.Status.Conditions, kapi.Condition{
+		Type:               kapi.ConditionSucceeded,
+		Status:             corev1.ConditionUnknown,
+		Reason:             "PipelineRunPending",
+		Message:            "PipelineRun is pending, waiting for Kueue admission",
```
pramodbindal: Lets say waiting for multikueue admission

internal/controller/pipelinerun_controller.go
ankrsinha: Made the changes, moved the reconciler into its own new file.

internal/controller/pipelinerun_controller.go
ankrsinha: Updated the Message field accordingly.

ankrsinha:
@pramodbindal I have made the changes as per your suggestions, please have a look.

internal/controller/pipelinerun_status_reconciler.go
```diff
@@ -0,0 +1,60 @@
...
+// condition with reason "PipelineRunPending" on runs that are suspended
+// (spec.status=PipelineRunPending) but have no status.conditions yet. Tekton's
+// controller skips PipelineRuns with spec.managedBy set, so without this
+// reconciler the status stays blank in CLI and UI tools until MultiKueue admits
+// the workload and dispatches it to a spoke cluster.
+type pipelineRunStatusReconciler struct {
+	client client.Client
+}
+
+func (r *pipelineRunStatusReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
gbenhaim, line 29: Can we add this logic to the existing controller https://github.com/konflux-ci/tekton-kueue/blob/main/internal/controller/pipelinerun_controller.go ?

There are many hook function that kueue's job framework provides, can we use any of those ?

gbenhaim (changes requested):

internal/controller/pipelinerun_status_reconciler.go
ankrsinha: @gbenhaim I checked the available hook interfaces in kueue's job framework. The hooks that have client access (such as Stop, Finalize, and GetCustomAnnotations) are only invoked during eviction, finalization, or workload slicing. 

They aren't called while a `PipelineRun` is in pending / yet to be admitted state, which is when we need to update the` status` condition. So I don't think the existing hooks can handle this use case.

pramodbindal: @ankrsinha  please add the unit tests

pramodbindal: @gbenhaim 
Here we are  updating the pipeline which suspended and will be picked by one of the spoke clusters.

I think adding a dedicated reconciler is right thing to do

ankrsinha: Verified by adding debug logs to existing hooks, that the hooks that get called are `PodSets`, `Finished`, `IsSuspended` and `isActive`, but these are pure data methods with no client access, so cannot make API calls.
Screenshot From 2026-07-30 15-24-28

ankrsinha: @pramodbindal Added the unit tests.

pramodbindal (approved):

internal/controller/pipelinerun_status_reconciler.go
pramodbindal: @gbenhaim  please  review again

internal/controller/pipelinerun_status_reconciler.go
```diff
@@ -0,0 +1,60 @@
...
+
+	return ctrl.Result{}, r.client.Status().Update(ctx, &plr)
+}
+
+// SetupPendingStatusWithManager registers the pending-status reconciler with the manager.
+func SetupPendingStatusWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		Named("PipelineRunStatus").
+		For(&tekv1.PipelineRun{}).
+		Complete(&pipelineRunStatusReconciler{client: mgr.GetClient()})
```
gbenhaim, line 80: **Suggestion**: The controller watches all PipelineRuns without filtering. In clusters with many PipelineRuns, every event triggers a reconcile that immediately short-circuits. Extract the filter logic into a standalone predicate function so it can be unit tested, then wire it in:

```go
// needsPendingStatusCondition returns true for PipelineRuns that are
// MultiKueue-managed, pending, and likely missing a status condition.
func needsPendingStatusCondition(o client.Object) bool {
	plr, ok := o.(*tekv1.PipelineRun)
	if !ok {
		return false
	}
	return plr.Spec.ManagedBy != nil &&
		*plr.Spec.ManagedBy == common.ManagedByMultiKueueLabel &&
		plr.Spec.Status == tekv1.PipelineRunSpecStatusPending
}

func SetupPendingStatusWithManager(mgr ctrl.Manager) error {
	return ctrl.NewControllerManagedBy(mgr).
		Named("PipelineRunStatus").
		For(&tekv1.PipelineRun{}, builder.WithPredicates(
			predicate.NewPredicateFuncs(needsPendingStatusCondition),
		)).
		Complete(&pipelineRunStatusReconciler{client: mgr.GetClient()})
}
```

This avoids enqueuing work items that will always no-op, and the extracted function can be tested independently.

internal/controller/pipelinerun_controller.go
```diff
@@ -104,7 +106,11 @@ func SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
 		return err
 	}
 
-	return reconciler.SetupWithManager(mgr)
+	if err := reconciler.SetupWithManager(mgr); err != nil {
+		return err
+	}
+
+	return SetupPendingStatusWithManager(mgr)
```
gbenhaim: **Suggestion**: Following kubebuilder conventions, each controller should be registered independently from the entrypoint (`cmd/main.go`), rather than having one controller register another. This makes the lifecycle explicit and consistent with how `SetupIndexer` and the `ConfigMapReconciler` are already registered separately.

Move `SetupPendingStatusWithManager(mgr)` to `cmd/main.go`:

```go
// cmd/main.go
err = controller.SetupWithManager(ctx, mgr)
// ...
if err = controller.SetupPendingStatusWithManager(mgr); err != nil {
	setupLog.Error(err, "Failed to setup pending status controller")
	os.Exit(1)
}
```

And revert `SetupWithManager` to return directly:
```go
return reconciler.SetupWithManager(mgr)
```

internal/controller/pipelinerun_status_reconciler.go
```diff
@@ -0,0 +1,60 @@
...
+
+	if plr.Spec.Status != tekv1.PipelineRunSpecStatusPending {
+		return ctrl.Result{}, nil
+	}
+
+	if plr.Status.GetCondition(kapi.ConditionSucceeded) != nil {
+		return ctrl.Result{}, nil
+	}
+
+	plr.Status.Conditions = append(plr.Status.Conditions, kapi.Condition{
```
gbenhaim: **Nit**: `PipelineRunStatus` has a `SetCondition` method that handles deduplication and keeps conditions sorted. While the `GetCondition != nil` guard above prevents duplicates, using `SetCondition` is more defensive and idiomatic with the Tekton API:

```go
plr.Status.SetCondition(&kapi.Condition{
	Type:               kapi.ConditionSucceeded,
	Status:             corev1.ConditionUnknown,
	Reason:             "PipelineRunPending",
	Message:            "PipelineRun is pending, waiting for MultiKueue admission",
	LastTransitionTime: kapi.VolatileTime{Inner: metav1.Now()},
})
```

internal/controller/pipelinerun_status_reconciler.go
gbenhaim, line 29: **Nit**: Consider adding a log line when the condition is actually written. This helps operators confirm the reconciler is working and trace which PipelineRuns it touched:

```go
func (r *pipelineRunStatusReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
	log := ctrl.LoggerFrom(ctx)
	// ... existing guards ...

	log.V(2).Info("Setting PipelineRunPending status condition")
	return ctrl.Result{}, r.client.Status().Update(ctx, &plr)
}
```

gbenhaim (commented):
_This review was posted by **Gal's Cursor**._

internal/controller/pipelinerun_status_reconciler.go
ankrsinha: Switched to `SetCondition` as suggested.

internal/controller/pipelinerun_status_reconciler.go
ankrsinha: Added `needsPendingStatusCondition` predicate with unit tests so only matching PipelineRuns are enqueued.

internal/controller/pipelinerun_status_reconciler.go
ankrsinha: Added log before the status update.

internal/controller/pipelinerun_controller.go
ankrsinha: Moved the controller registration to c`md/main.go`, `SetupWithManager `now returns directly (reverted all the changes in `pipelinerun_controller.go`).

ankrsinha:
@gbenhaim I have updated the PR with all the changes suggested in the review comments.

gbenhaim (approved):

PR #513, opened 2026-07-26
feat(cel): add managedBy() and multiKueue() CEL functions for routing

khrm:
## Verification

- [x] `make lint` passes
- [x] `make test` passes
- [x] Tested with `tekton-kueue mutate` CLI (if CEL/webhook changes)
- [x] `make manifests` and `make generate` run cleanly (if API/RBAC changes)

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,32 @@ data:
 ```
 The controller will automatically load the configuration from this `ConfigMap`.
 
+#### CEL-Based Routing (Alternative)
+
+For conditional per-PipelineRun routing, use the CEL `multiKueue()` function instead of the boolean `multiKueueOverride`. This lets you route only specific PipelineRuns to MultiKueue based on labels, namespace, or other properties:
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: tekton-kueue-config
+  namespace: tekton-kueue-system
+data:
+  config.yaml: |
+    queueName: default-queue
+    cel:
+      expressions:
+        # Route build pipelines to spoke cluster via MultiKueue,
+        # keep everything else on the hub
+        - |
+          has(pipelineRun.metadata.labels) &&
+          "pipeline-type" in pipelineRun.metadata.labels &&
+          pipelineRun.metadata.labels["pipeline-type"] == "build" ?
+            [multiKueue("multikueue-queue")] : []
+```
+
+> **Note:** When using CEL routing functions (`managedBy()`, `multiKueue()`), do not set `multiKueueOverride: true`. The boolean sets `Spec.ManagedBy` for all PipelineRuns unconditionally *before* CEL expressions are evaluated, which prevents conditional per-PipelineRun routing. Use `multiKueueOverride` only when *all* PipelineRuns should be routed to MultiKueue.
+
 ## Command Line Interface
 
 The `tekton-kueue` binary provides several subcommands:
@@ -260,6 +286,25 @@ cel:
     - 'plrNamespace == "production" ? label("environment", "prod") : label("environment", "dev")'
     - 'pacEventType == "push" ? annotation("trigger", "push-event") : annotation("trigger", "other-event")'
     - 'pacTestEventType != "" ? label("test-type", pacTestEventType) : label("test-type", "none")'
+    
+    # managedBy - set Spec.ManagedBy to a custom controller (standalone example)
+    - 'managedBy("tekton.dev/pipeline")'
+    
+    # multiKueue - route to spoke cluster via MultiKueue (standalone example)
+    - 'multiKueue()'
+    
+    # multiKueue with queue - route to spoke AND override queue name (standalone example)
+    - 'multiKueue("multikueue-queue")'
+```
+
+> **Note:** `managedBy()`, `multiKueue()`, and `multiKueue("queueName")` all set `Spec.ManagedBy`. Using more than one unconditionally in the same config does not make sense — the last one silently overrides the others. Use them in conditional expressions (ternaries) for per-PipelineRun routing, or pick a single routing strategy.
+
+```yaml
+cel:
+  expressions:
+    # Conditional routing - builds to spoke, releases stay on hub
+    - 'has(pipelineRun.metadata.labels) && "pipeline-type" in pipelineRun.metadata.labels && pipelineRun.metadata.labels["pipeline-type"] == "build" ? [multiKueue("multikueue-queue")] : []'
+    - 'has(pipelineRun.metadata.labels) && "pipeline-type" in pipelineRun.metadata.labels && pipelineRun.metadata.labels["pipeline-type"] == "release" ? [managedBy("tekton.dev/pipeline")] : []'
     
     # Multiline CEL expression for multiple mutations
     # This expression applies several annotations and labels in one go
--- a/internal/cel/compiler.go
+++ b/internal/cel/compiler.go
@@ -58,6 +58,8 @@ func createCELEnvironment() (*cel.Env, error) {
 		createMutationFunction("label", MutationTypeLabel, mutationRequestType),
 		createResourceMutationFunction("resource", MutationTypeResource, mutationRequestType),
 		createPriorityMutationFunction("priority", mutationRequestType),
+		createManagedByFunction("managedBy", mutationRequestType),
+		createMultiKueueFunction("multiKueue", mutationRequestType),
 		// Add string manipulation functions
 		createReplaceFunction("replace"),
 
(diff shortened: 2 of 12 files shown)


internal/cel/mutator_test.go
```diff
@@ -851,6 +948,113 @@ func TestCELMutator_Mutate(t *testing.T) {
...
+					Labels:    maps.Clone(tt.initialLabels),
+				},
+				Spec: tekv1.PipelineRunSpec{
+					PipelineRef: &tekv1.PipelineRef{Name: "test-pipeline"},
+				},
+			}
+
+			programs, err := CompileCELPrograms(tt.expressions)
+			g.Expect(err).NotTo(HaveOccurred())
+
```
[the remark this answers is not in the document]
khrm: This is a false positive.  We are using this pattern in other places also.

khrm (commented):
/assign @gbenhaim 

Please review and merge

internal/cel/doc.go
```diff
@@ -12,7 +12,7 @@
 //
 //   - Input: *tekton.PipelineRun (strongly typed and validated)
 //   - Output: []MutationRequest (validated structure and content)
-//   - Functions: annotation(key, value), label(key, value), and priority(value)
+//   - Functions: annotation(), label(), priority(), resource(), managedBy(), multiKueue(), replace()
 //   - Expressions: Single mutations or lists of mutations
```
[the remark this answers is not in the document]
khrm: Fixed.

gbenhaim (commented):
_This review was posted by **Gal's Cursor**._

README.md
```diff
@@ -261,6 +261,19 @@ cel:
...
+    
+    # multiKueue - route to spoke cluster via MultiKueue
+    - 'multiKueue()'
+    
+    # multiKueue with queue - route to spoke AND override queue name
+    - 'multiKueue("multikueue-queue")'
+    
+    # Conditional routing - builds to spoke, releases stay on hub
+    - 'has(pipelineRun.metadata.labels) && "pipeline-type" in pipelineRun.metadata.labels && pipelineRun.metadata.labels["pipeline-type"] == "build" ? multiKueue("multikueue-queue") : []'
+    - 'has(pipelineRun.metadata.labels) && "pipeline-type" in pipelineRun.metadata.labels && pipelineRun.metadata.labels["pipeline-type"] == "release" ? [managedBy("tekton.dev/pipeline")] : []'
```
gbenhaim, line 307: **Suggestion — clarify interaction between `multiKueueOverride` and CEL routing functions**

The new CEL routing functions (`managedBy()`, `multiKueue()`) and the existing `multiKueueOverride: true` config both set `Spec.ManagedBy`, but the config boolean runs first (before CEL mutators). This means:

- If `multiKueueOverride: true` is set, `ManagedBy` is set for **all** PipelineRuns before CEL expressions evaluate. A conditional CEL expression like `condition ? managedBy("custom") : []` would only override the true branch — the false branch still has `ManagedBy` set from the config.
- CEL `managedBy()` / `multiKueue()` can override the config-level value, but not unset it.

Consider adding a note here (or in the MultiKueue section above) along the lines of:

> **Note:** When using CEL routing functions (`managedBy()`, `multiKueue()`), do not set `multiKueueOverride: true` in the config. The boolean sets `Spec.ManagedBy` for all PipelineRuns unconditionally before CEL expressions are evaluated, which prevents conditional per-PipelineRun routing. Use `multiKueueOverride` only when all PipelineRuns should be routed to MultiKueue.

gbenhaim (commented):
_This review was posted by **Gal's Cursor**._

internal/cel/compiler_test.go
```diff
@@ -34,6 +34,10 @@ func TestCompileCELPrograms_TypeSafety(t *testing.T) {
 				`resource("example.com/cpu", 1000)`,
 				`[annotation("key1", "value1"), label("key2", "value2")]`,
 				`priority("high")`,
+				`managedBy("custom-controller.io")`,
+				`multiKueue()`,
+				`multiKueue("mk-queue")`,
+				`[multiKueue(), annotation("k", "v")]`,
```
gbenhaim, line 40: **Suggestion — add a compile-failure test for `condition ? multiKueue() : multiKueue("queue")`**

Since the two `multiKueue` overloads have different return types (`map` vs `list`), mixing them in a ternary is a compile-time error. Adding a negative test case in the error section below would document this behavior and guard against future changes that might accidentally alter the type resolution:

```go
{
    name: "type error - multiKueue overloads mixed in ternary",
    expressions: []string{
        `true ? multiKueue() : multiKueue("queue")`,
    },
    expectErr: true,
},
```

Also worth adding: a webhook-level test in `pipelinerun_webhook_test.go` verifying that CEL `managedBy("custom")` overrides the config-level `multiKueueOverride: true` — this would document the intended precedence.

internal/cel/types_test.go
```diff
@@ -142,6 +142,25 @@ func TestMutationRequest_Validate(t *testing.T) {
 			expectErr: true,
 			errMsg:    "mutation value cannot be empty",
 		},
+		{
```
gbenhaim, line 159: **Suggestion — also add `MutationTypeManagedBy` to `TestMutationType_IsValid` and `TestMutationType_JSON` above**

The new `managedBy` type is covered here in `Validate`, but the `IsValid` and `JSON` test tables above don't include it. Adding explicit cases would improve completeness and catch regressions if someone removes the type from the enum:

```go
// In TestMutationType_IsValid:
{"valid managedBy", MutationTypeManagedBy, true},

// In TestMutationType_JSON:
{
    name:      "valid managedBy",
    input:     `"managedBy"`,
    expectErr: false,
    expected:  MutationTypeManagedBy,
},
```

gbenhaim (commented):
_This review was posted by **Gal's Cursor**._

internal/cel/mutator.go
```diff
@@ -161,6 +162,8 @@ func mutate(pipelineRun *tekv1.PipelineRun, mutation *MutationRequest) (*tekv1.P
 
 		// Store the summed value back as string
 		pipelineRun.Annotations[mutation.Key] = strconv.Itoa(newValue)
+	case MutationTypeManagedBy:
+		pipelineRun.Spec.ManagedBy = ptr.To(mutation.Value)
```
gbenhaim, line 167: **Suggestion — add e2e and/or patch-filter test coverage for `ManagedBy` mutations**

The `MutationTypeManagedBy` case here is well-covered by unit tests, but there's a gap in integration coverage:

1. **No patch-filter test for `/spec/managedBy`**: The webhook unit tests in `pipelinerun_webhook_test.go` test the patch filter (`patchFilteringWebhook.Handle()`) only with a basic `QueueName` config. There's no test verifying that a `managedBy` patch actually survives the filter — the `/spec/managedBy` entry in `allowedPatchPrefixes` is never exercised. The existing `multiKueueOverride: true` test calls `Default()` directly, bypassing the patch filter entirely.

2. **No e2e test for the full admission round-trip**: A lightweight e2e test following the existing ConfigMap reload pattern could verify that `Spec.ManagedBy` survives the full webhook admission path (patch filtering + Tekton's own webhook interaction). Something like:

```go
Context("CEL managedBy sets Spec.ManagedBy", Ordered, func() {
    It("should set ManagedBy on created PipelineRun", func(ctx context.Context) {
        // Update ConfigMap with CEL expression: managedBy("test-controller")
        // Create PipelineRun
        // Verify plr.Spec.ManagedBy == "test-controller"
    })
})
```

At minimum, adding a patch-filter unit test with a `managedBy()` CEL expression would close the gap without requiring e2e infrastructure changes.

gbenhaim (commented):
_This review was posted by **Gal's Cursor**._

internal/cel/compiler.go
```diff
@@ -217,6 +220,89 @@ func createPriorityMutationFunction(name string, returnType *cel.Type) cel.EnvOp
...
+					"key":   "managedBy",
+					"value": common.ManagedByMultiKueueLabel,
+				}
+				return types.NewStringInterfaceMap(types.DefaultTypeAdapter, mutationMap)
+			}),
+		),
+		cel.Overload(
+			name+"_string_to_mutation_list",
+			[]*cel.Type{cel.StringType},
+			listReturnType,
```
gbenhaim: **Concern — `multiKueue("queue")` is the only CEL function returning a list, causing type incompatibilities**

Every other CEL function (`annotation()`, `label()`, `priority()`, `resource()`, `managedBy()`, `multiKueue()` zero-arg) returns `map<string, any>`. The one-arg `multiKueue("queue")` is the only one that returns `list<map<string, any>>`. This creates several issues:

1. **Can't mix overloads in ternaries** — `condition ? multiKueue() : multiKueue("queue")` fails to compile because the branches have different types.
2. **Can't compose in lists** — `[multiKueue("queue"), annotation("k", "v")]` fails because you'd get a list containing a list and a map.
3. **Surprising to users** — every other function returns a single mutation; this one is a special case.

**Suggested approach: return a single map with a dedicated mutation type**

Have both `multiKueue()` overloads return `map<string, any>` with a new `MutationTypeMultiKueue` type. The queue name goes in the `value` field (empty string for zero-arg). Then `mutate()` in `mutator.go` handles expanding it:

```go
// compiler.go — both overloads return the same type (returnType, not listReturnType)
cel.Overload(
    name+"_to_mutation",
    []*cel.Type{},
    returnType,
    cel.FunctionBinding(func(args ...ref.Val) ref.Val {
        return types.NewStringInterfaceMap(types.DefaultTypeAdapter, map[string]interface{}{
            "type":  string(MutationTypeMultiKueue),
            "key":   "managedBy",
            "value": "",  // no queue override
        })
    }),
),
cel.Overload(
    name+"_string_to_mutation",
    []*cel.Type{cel.StringType},
    returnType,  // same type as zero-arg!
    cel.UnaryBinding(func(val ref.Val) ref.Val {
        // ... validate queueName ...
        return types.NewStringInterfaceMap(types.DefaultTypeAdapter, map[string]interface{}{
            "type":  string(MutationTypeMultiKueue),
            "key":   "managedBy",
            "value": queueName,
        })
    }),
),
```

```go
// mutator.go — expand the composite mutation at apply time
case MutationTypeMultiKueue:
    pipelineRun.Spec.ManagedBy = ptr.To(common.ManagedByMultiKueueLabel)
    if mutation.Value != "" {
        if pipelineRun.Labels == nil {
            pipelineRun.Labels = make(map[string]string)
        }
        pipelineRun.Labels[common.QueueLabel] = mutation.Value
    }
```

This keeps all CEL functions returning the same type, makes both overloads interchangeable in ternaries and lists, and moves the "two operations from one mutation" logic to `mutate()` where it's a natural fit.

gbenhaim (commented):
_This review was posted by **Gal's Cursor**._

gbenhaim (commented):
_This review was posted by **Gal's Cursor**._

README.md
```diff
@@ -261,6 +261,19 @@ cel:
...
     - 'pacTestEventType != "" ? label("test-type", pacTestEventType) : label("test-type", "none")'
     
+    # managedBy - set Spec.ManagedBy to a custom controller
+    - 'managedBy("tekton.dev/pipeline")'
+    
+    # multiKueue - route to spoke cluster via MultiKueue
+    - 'multiKueue()'
+    
+    # multiKueue with queue - route to spoke AND override queue name
+    - 'multiKueue("multikueue-queue")'
```
gbenhaim, line 297: **Suggestion — improve README examples for routing functions**

1. **Misleading placement in the kitchen-sink example block**: The unconditional `managedBy("tekton.dev/pipeline")` (line 265) and `multiKueue()` (line 268) appear alongside all other examples in the same `expressions:` list. If someone copy-pastes this config, every PipelineRun would get both applied — which contradicts each other. These should have a comment clarifying they are standalone examples, not meant to be used together.

2. **Add a dedicated routing example to the "Usage with MultiKueue" section** (line 120 in the existing README): That section currently only documents the `multiKueueOverride: true` boolean approach. Since this PR introduces the CEL-based alternative, consider extending that section with a realistic conditional routing config example, e.g.:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: tekton-kueue-config
  namespace: tekton-kueue-system
data:
  config.yaml: |
    queueName: default-queue
    cel:
      expressions:
        # Route build pipelines to spoke cluster via MultiKueue,
        # keep everything else on the hub
        - |
          has(pipelineRun.metadata.labels) &&
          "pipeline-type" in pipelineRun.metadata.labels &&
          pipelineRun.metadata.labels["pipeline-type"] == "build" ?
            multiKueue("multikueue-queue") : []
```

This gives users a copy-pasteable starting point and makes the connection between the old `multiKueueOverride` boolean and the new CEL routing explicit.

khrm: I have added that. And also added a note/clarification on this.

internal/cel/compiler.go
khrm: Fixed.

README.md
khrm: Done.

gbenhaim (approved):

PR #520, opened 2026-07-27
WIP - Add MultiKueue Controller to create  cluster secrets for spoke clusters

pramodbindal:
Create Service Account on spoke cluster
Create Secret of type service-account-token
Create Role and Rolebinding for the secret so hub is able to access the Spoke Resources Create a Spoke Secret on hub cluster

## Verification

- [ ] `make lint` passes
- [ ] `make test` passes
- [ ] Tested with `tekton-kueue mutate` CLI (if CEL/webhook changes)
- [ ] `make manifests` and `make generate` run cleanly (if API/RBAC changes)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,7 +139,7 @@ build-installer: manifests generate kustomize ## Generate a consolidated YAML wi
 	$(KUSTOMIZE) build config/default > dist/install.yaml
 
 .PHONY: release
-release: kustomize
+release: kustomize manifests generate
 	mkdir -p ${RELEASE_DIR}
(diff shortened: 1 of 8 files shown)


gbenhaim:
Can you please add more context, maybe a design document which explain what is the purpose of this controller, which problems it solves, what is the authentication model, testing stredegy and why you choose to include it in tekton-kueue?

gbenhaim (changes requested):
We need more context and design for this change. Please see my comment above.

pramodbindal:
Logic is moved out of tekton-kueue

PR #521, opened 2026-07-30
Derive Tekton and Kueue versions from go.mod for e2e tests

gbenhaim:
## What

- Replace single-purpose `hack/get-kueue-version.sh` with generic `hack/get-dep-version.sh` that extracts any Go module version from `go.mod`
- Derive both `KUEUE_VERSION` and `TEKTON_VERSION` in the Makefile from `go.mod`, ensuring e2e tests always install the same versions the code compiles against
- Update `tektoncd/pipeline` from v1.11.1 to v1.12.2

## Why

`TEKTON_VERSION` was hardcoded and had drifted from the actual Go module dependency. This caused e2e tests to install a different Tekton version than what the code was built against. Deriving both versions from `go.mod` keeps them in sync automatically.

## Verification

- [x] `go build ./...` passes
- [x] `hack/get-dep-version.sh` correctly resolves both `sigs.k8s.io/kueue` and `github.com/tektoncd/pipeline` versions

Made with [Cursor](https://cursor.com)


Makefile
```diff
@@ -193,8 +193,8 @@ ENVTEST_VERSION ?= $(shell go list -m -f "{{ .Version }}" sigs.k8s.io/controller
 #ENVTEST_K8S_VERSION is the version of Kubernetes to use for setting up ENVTEST binaries (i.e. 1.31)
 ENVTEST_K8S_VERSION ?= $(shell go list -m -f "{{ .Version }}" k8s.io/api | awk -F'[v.]' '{printf "1.%d", $$3}')
 GOLANGCI_LINT_VERSION ?= v2.7.2
-KUEUE_VERSION ?= $(shell ./hack/get-kueue-version.sh)
-TEKTON_VERSION ?= v1.12.2
+KUEUE_VERSION ?= $(shell ./hack/get-dep-version.sh sigs.k8s.io/kueue)
```
khrm: Why not use 
```
KUEUE_VERSION ?= $(shell go list -m -f "{{ .Version }}" sigs.k8s.io/kueue)
TEKTON_VERSION ?= $(shell go list -m -f "{{ .Version }}" github.com/tektoncd/pipeline)
```
?

I see that we already use this pattern for ENVTEST_K8S_VERSION and ENVTEST_VERSION.

gbenhaim: good idea! done.

khrm (commented):
/approve

pramodbindal (approved):

PR #522, opened 2026-08-02
Fix: Pin kind version and node image in E2E workflow

meyrevived:
## Summary
- Pin kind to v0.32.0 (latest stable) instead of floating `/dl/latest/`
- Pin node image to `kindest/node:v1.35.5`
- Add Renovate annotations so both versions are kept up to date automatically

## Root cause
`kind.sigs.k8s.io/dl/latest/` resolved to `v0.33.0-alpha`, which ships `kindest/node:v1.36.1`. That node image uses an OCI runtime spec version incompatible with the `crun` binary on GitHub Actions ubuntu runners when `KIND_EXPERIMENTAL_PROVIDER=podman`, producing `crun: unknown version specified` (exit 126) at cluster creation.

## Approach
Rather than hardcoding a version that rots, the kind binary version and node image are annotated with Renovate datasource comments:
- **kind binary**: `datasource=github-releases depName=kubernetes-sigs/kind` — Renovate's default semver versioning excludes pre-release tags (`-alpha`, `-beta`, `-rc`), so only stable kind releases are proposed.
- **node image**: `datasource=docker depName=kindest/node` — tracked as a Docker image tag.

Because kind and its node image are not independently versionable, a Renovate update to one without the other could cause a mismatch. This is safe because **CI creates a kind cluster on every PR** — an incompatible pairing fails the E2E workflow before merge, making the mismatch visible and blocking the update.

## Test plan
- [ ] E2E tests pass with pinned versions (CI validates this)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -20,9 +20,9 @@ jobs:
         with:
           go-version-file: go.mod
 
-      - name: Install the latest version of kind
+      - name: Install pinned kind version
         run: |
-          curl -Lo ./kind https://kind.sigs.k8s.io/dl/latest/kind-linux-amd64
+          curl -Lo ./kind https://kind.sigs.k8s.io/dl/v0.32.0/kind-linux-amd64
           chmod +x ./kind
           sudo mv ./kind /usr/local/bin/kind
 
@@ -30,7 +30,7 @@ jobs:
         run: kind version
 
       - name: Create kind cluster
-        run: kind create cluster
+        run: kind create cluster --image kindest/node:v1.35.5
         env:
           KIND_EXPERIMENTAL_PROVIDER: podman
 
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -21,8 +21,13 @@ jobs:
           go-version-file: go.mod
 
       - name: Install pinned kind version
+        # Renovate updates KIND_VERSION automatically; only stable releases are proposed.
+        # If kind and node image are updated independently a version mismatch may cause
+        # cluster creation to fail, which CI will catch before merge.
         run: |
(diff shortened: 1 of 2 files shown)


ilaydamari (approved):

khrm (commented):
/lgtm

glevi-rh (approved):

.github/workflows/test-e2e.yml
```diff
@@ -20,17 +20,22 @@ jobs:
...
+          curl -Lo ./kind "https://kind.sigs.k8s.io/dl/${KIND_VERSION}/kind-linux-amd64"
           chmod +x ./kind
           sudo mv ./kind /usr/local/bin/kind
 
       - name: Verify kind installation
         run: kind version
 
       - name: Create kind cluster
-        run: kind create cluster
+        run: kind create cluster --image kindest/node:v1.35.5 # renovate: datasource=docker depName=kindest/node versioning=docker
```
filariow: kind `v0.32.0` is using `kindest/node:v1.36.0`. Is there a specific reason for us to fix it to `v1.35.5`?
Should we manage `kind`'s version only?

meyrevived: From the PR's description under "Root Cause":
`kind.sigs.k8s.io/dl/latest/` resolved to `v0.33.0-alpha`, which ships `kindest/node:v1.36.1`. That node image uses an OCI runtime spec version incompatible with the crun binary on GitHub Actions ubuntu runners when `KIND_EXPERIMENTAL_PROVIDER=podman`, producing `crun: unknown version specified (exit 126)` at cluster creation.

sadlerap: would it be possible to pin the hashes of the node images we use while still having renovate manage them?

meyrevived: Done in latest PR

oswcab (approved):

PR #530, opened 2026-08-10
build: prepare reproducible release artifacts

waveywaves:
# Changes

- Move default deployment and release images to `ghcr.io/tektoncd/tekton-kueue` and add a digest-pinned `ko` base image.
- Render manifests through a temporary Kustomize overlay so release and deploy commands no longer modify tracked configuration.
- Require immutable images for local releases, generate checksums, and add `make verify-release`.
- Remove the pre-transfer Pipelines as Code definitions that would conflict with Tekton's release controller.

The Go module path and existing `kueue.konflux-ci.dev` API keys remain unchanged for compatibility.

Part of #527.

/kind feature

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) if any changes are user facing, including updates to minimum requirements e.g. Kubernetes version bumps
- [x] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including functionality, content, code)
- [x] Has a kind label. You can add one by adding a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [x] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings). See some examples of [good release notes](https://github.com/tektoncd/community/blob/main/standards.md#good-release-notes).
- [x] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
Release manifests use digest-qualified images from ghcr.io/tektoncd/tekton-kueue.
```

PR #529, opened 2026-08-10
docs: define release policy and runbook

waveywaves:
# Changes

- Propose a quarterly LTS cadence beginning with `v0.5.0` and document the support boundary required by the Tekton release policy.
- Define the intended Tekton-controlled image, manifest, checksum, signing, and provenance contract.
- Add an operator runbook while leaving release-branch naming to the resolution of #20.

Part of #527.

/kind documentation

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) if any changes are user facing, including updates to minimum requirements e.g. Kubernetes version bumps
- [x] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including functionality, content, code)
- [x] Has a kind label. You can add one by adding a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [x] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings). See some examples of [good release notes](https://github.com/tektoncd/community/blob/main/standards.md#good-release-notes).
- [x] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
NONE
```

Issue #527, opened 2026-08-07
Complete Tekton project integration and release readiness


waveywaves:
## Goal

Complete the post-adoption work needed for `tekton-kueue` to operate like other top-level Tekton projects and to publish official, reproducible Tekton releases.

The repository transfer is complete and [TEP-0164](https://github.com/tektoncd/community/blob/main/teps/0164-tekton-kueue-integration.md) already identifies CI/CD, release automation, registry publication, and documentation as adoption work. Today the repository has tests, `OWNERS`, protected release branches, and historical GitHub releases, but the standard Tekton project/release integration is incomplete.

## Project and community integration

- [ ] Finish the review/status transition for TEP-0164 and update its repository, registry, compatibility, and implementation links.
- [ ] Add the required project documents from the [Tekton project requirements](https://github.com/tektoncd/community/blob/main/process/projects.md#project-requirements):
  - [ ] `DEVELOPMENT.md`
  - [ ] `CONTRIBUTING.md`
  - [ ] Tekton Code of Conduct reference/copy
- [ ] Replace the current PR template with the standard Tekton template, including contributor-standard checks, kind labeling, and a `release-note` block.
- [ ] Add or explicitly enable the standard Tekton bug, feature, and free-form issue templates from [`tektoncd/.github`](https://github.com/tektoncd/.github/tree/main/.github), with Kueue/Pipelines version diagnostics where useful.
- [ ] Align `OWNERS`/`OWNERS_ALIASES` with the Tekton Kueue maintainer and collaborator teams and define both approvers and reviewers.
- [ ] Add the standard Tekton/Prow labels and automation (`kind/*`, release-note labels, merge-blocking labels, `approved`, and `lgtm`).
- [ ] Remove stale pre-transfer identity and publication references from `PROJECT`, `.tekton/`, README, default image coordinates, generated manifests, and contributor guidance.
- [ ] Set repository topics/homepage and add installation/user documentation to the appropriate Tekton documentation surface.

## CI and repository automation

- [ ] Add `tektoncd/tekton-kueue` to Tekton plumbing/Tide, including `config/repo-checks.yaml`, merge configuration, and Terraform-managed branch protection.
- [ ] Provide one stable required `CI summary` check covering build/generation, lint, unit tests, and e2e tests; keep all actions and tools pinned.
- [ ] Verify `main`, supported release branches, and release tags prevent force-push/deletion and require the expected reviews/checks.
[...]

Issue #20, opened 2025-04-24
Use release branches


gbenhaim:
Follow the branch schema kueue is using for ensuring tekton-kueue is compatiable with kueue.
We should start using release branches and follow the same versions as kueue. So for example, release branch release-0.x on tekton kueue will be compatible with the same branch on kueue.

PR #531, opened 2026-08-10
ci: add Tekton release automation

waveywaves:
# Changes

- Add trusted Pipelines as Code entry points for initial and patch releases.
- Build one four-platform controller image in `ghcr.io/tektoncd/tekton-kueue`, generate tagged and digest-only manifests plus checksums, and expose the image to Tekton Chains.
- Upload versioned and latest artifacts, wait for signing and provenance, and create a GitHub draft release for operator review.
- Add scheduled and manual patch-release discovery with branch/version validation and immutable executable references.

This is stacked on #530; I will rebase it to leave only the release-automation commit after that prerequisite merges. The release policy and operator runbook are proposed in #529. Initial and patch automation accepts the existing `.x` convention and Tekton's `-lts` convention while #20 is resolved.

Part of #527.

/kind feature

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) if any changes are user facing, including updates to minimum requirements e.g. Kubernetes version bumps
- [x] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including functionality, content, code)
- [x] Has a kind label. You can add one by adding a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [x] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings). See some examples of [good release notes](https://github.com/tektoncd/community/blob/main/standards.md#good-release-notes).
- [x] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
NONE
```


hack/render-manifest.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
```
pramodbindal, line 1: we are doing the same thing from Makefile as well.

shouldn;t we remove it from there if this case ?

pramodbindal (approved):

hack/render-manifest.sh
waveywaves: The Makefile only calls this script. The rendering logic is not duplicated there.

waveywaves:
Rebased onto current `main` and hardened the release flow.

Key fixes:

- Use the correct bucket/project path during prerelease checks.
- Resolve the in-toto Rekor attestation instead of the per-image signature entry.
- Use the current pinned GitHub release task while keeping draft publication explicit for operator review.
- Validate incoming release parameters and pass them to shell steps through environment variables.
- Build webhook payloads with `jq` rather than shell interpolation.
- Prevent an external `KUSTOMIZE` binary from being rebuilt or replaced.
- Run `make verify-release` in CI.

Validation completed:

```text
make test
make verify-release
go build ./...
go mod tidy -diff
golangci-lint run ./...
actionlint
yamllint
Tekton v1 API validation
four-platform local OCI image/manifest resolution
scheduled/manual release workflow harnesses
```

The generated GitHub release intentionally remains a draft until an operator verifies the artifacts and signing data. Fresh CI is running on `d2e04fc`.

pramodbindal (approved):

